PR #30, opened 2016-10-18
Volume operations and flows

shtripat:
@r0h4n @nthomas-redhat @nnDarshan please review this patch as its updated and works fine.


tendrl/gluster_bridge/atoms/volume/create.py
```diff
@@ -3,6 +3,8 @@
 
 class Create(object):
     def start(self, name, bricks):
-        subprocess.call(['gluster', 'volume', 'create',
-                         name, ' '.join(bricks), ' force'])
+        cmd = ['gluster', 'volume', 'create', name]
+        cmd.extend(bricks)
+        cmd.append('force')
+        subprocess.call(cmd)
```
TimothyAsir: Why gluster volume start function contain has create functionality?

tendrl/gluster_bridge/manager/rpc.py
```diff
@@ -45,6 +45,7 @@ def _acceptor(self):
                         break
                     except Exception as ex:
                         LOG.error(ex)
+            gevent.sleep(2)
```
TimothyAsir: Why sleep here!. May be you can give a comment note here.

etc/sds_operations_gluster-3.8.3.yaml
```diff
@@ -11,81 +11,432 @@ object_details:
     attrs:
       volname:
         type: String
+	value: etcd
+	help: Gluster volume name
```
TimothyAsir: Remove the tabs. \t are not allowed for yaml

etc/sds_operations_gluster-3.8.3.yaml
```diff
@@ -101,48 +452,228 @@ object_details:
 
   peer:
     attrs:
-    - hostname
+      hostname
```
TimothyAsir: ":" is missing for hostname

etc/sds_operations_gluster-3.8.3.yaml
```diff
@@ -101,48 +452,228 @@ object_details:
...
+	run: "tendrl.gluster_bridge.atoms.snapshot.config"
         type: "update"
+	conditions:
+	  pre_condition: None
+	  post_condition: None
+	name: "config_snapshot"
+	outputs:
+	  - "success"
+	global_inputs: None
+	local_inputs
```
TimothyAsir: ':' is missing

tendrl/gluster_bridge/atoms/volume/create.py
shtripat: This atom is for create volume only. But starts the volume after creation.

etc/sds_operations_gluster-3.8.3.yaml
shtripat: Will do that.

etc/sds_operations_gluster-3.8.3.yaml
```diff
@@ -11,81 +11,432 @@ object_details:
...
-
-          global_inputs:
-          - An immutable input_name => value dictionary which specifies any initial inputs that should be automatically injected into the atoms scope before the atom execution commences (this allows for providing atom local values that do not need to be provided by other atoms/dependents).
+            - "volname"
+          global_inputs: None
           local_inputs:
-          - A set or list of required inputs for this atom’s execute method
-          version: An immutable version that associates version information with this atom. It can be useful in resuming older versions of atoms. Standard major, minor versioning concepts should apply
+            - "volname"
+            - ["stripe_count"]
```
shtripat: Actually wanted to add an optional parameter. Not sure how to ?

etc/sds_operations_gluster-3.8.3.yaml
shtripat: Will do that

etc/sds_operations_gluster-3.8.3.yaml
shtripat: Will add

shtripat:
@TimothyAsir @nnDarshan @nthomas-redhat please review the patches. Incorporated the comments from Tim.

shtripat:
@aravindavk @atinmu please review.

tendrl/gluster_bridge/flows/create_volume.py
```diff
@@ -11,10 +11,50 @@ def __init__(self, api_job):
         self.atom = Create
 
     def start(self):
-        attributes = json.loads(self.api_job['attributes'].decode('utf-8'))
+        attributes = self.api_job['attributes']
+        print attributes
```
nnDarshan: please remove debug messages in this file.

tendrl/gluster_bridge/flows/create_volume.py
```diff
@@ -11,10 +11,50 @@ def __init__(self, api_job):
...
+        elif v_replica_count is not None:
+            print v_replica_count
+            self.atom().start(
+                vol_name,
+                brickdetails,
+                transport = v_transport,
+                replica_count = v_replica_count,
+                arbiter_count = v_arbiter_count
+            )
+            print "Hell"
```
nnDarshan: was it frustrating :D

tendrl/gluster_bridge/flows/add_volume_brick.py
```diff
@@ -0,0 +1,20 @@
+import json
+
+import etcd
+from tendrl.gluster_bridge.atoms.volume.add_brick import AddBrick
+
+
+class AddVolumeBrick(object):
+    def __init__(self, api_job):
+        super(AddVolumeBrick, self).__init__()
```
nnDarshan: not needed I guess. (in all the constructors of flows)

tendrl/gluster_bridge/flows/add_volume_brick.py
```diff
@@ -0,0 +1,20 @@
...
+        self.atom = AddBrick
+
+    def start(self):
+        attributes = self.api_job['attributes']
+        vol_name = attributes['volname']
+        bricks = attributes['brickdetails']
+        self.atom().start(vol_name, bricks)
+        self.api_job['status'] = "finished"
+        etcd.Client().write(self.api_job['request_id'],
+                            json.dumps(self.api_job))
```
nnDarshan: I feel you can have the above 2 lines in rpc.py, in such a way that if flow succeeds you can have status as Finished else status as Failed. Thus you avoid having this in every flow file.

tendrl/gluster_bridge/atoms/peer/detach.py
```diff
@@ -0,0 +1,7 @@
+import subprocess
+
+
+class Detach(object):
+    def start(self, peer):
+        subprocess.call(['gluster', 'peer', 'detach',
```
shtripat: I think this we would take up in a separate PR and yes for sure its there back of my mind. Thanks for suggestion :)

tendrl/gluster_bridge/atoms/volume/create.py
```diff
@@ -2,7 +2,44 @@
...
+        replica_count = None,
+        arbiter_count = None,
+        stripe_count = None,
+        disperse_count = None,
+        disperse_data_count = None,
+        redundancy_count = None,
+        transport = [],
+        ):
+        cmd = ['gluster', 'volume', 'create', name]
+        print replica_count
```
shtripat: Ah somehow forgot to remove. Done.

tendrl/gluster_bridge/atoms/volume/create.py
```diff
@@ -2,7 +2,44 @@
...
+                cmd.append('disperse_data')
+                cmd.append(disperse_data_count)
+        if transport:
+            cmd.append('transport')
+            cmd.append(','.join(transport))
+        cmd.extend(bricks)
+        cmd.append('force')
+        cmd.append('--mode=script')
+        subprocess.call(cmd)
+        subprocess.call(['gluster', 'volume', 'start',
```
shtripat: Thats good idea. @r0h4n @nthomas-redhat what you say? I am +1 for this.

tendrl/gluster_bridge/atoms/volume/delete.py
```diff
@@ -0,0 +1,9 @@
+import subprocess
+
+
+class Delete(object):
+    def start(self, name):
+        subprocess.call(['gluster', 'volume', 'stop',
+                        name, '--mode=script'])
+        subprocess.call(['gluster', 'volume', 'delete',
```
shtripat: +1

tendrl/gluster_bridge/flows/create_volume.py
shtripat: Haha. yeah will remove this ;)

tendrl/gluster_bridge/flows/add_volume_brick.py
shtripat: oh ok. will try that. just retained as was there in original code of create volume flow.

tendrl/gluster_bridge/flows/add_volume_brick.py
shtripat: good suggestion. need to try in another PR may be with return value check as well.

shtripat:
@brainfunked @r0h4n @nthomas-redhat @anmolbabu @nnDarshan @TimothyAsir I have incorporated the changes guidelines for spec yaml file. Kindly review.

etc/sds_operations_gluster-3.8.3.yaml
```diff
@@ -1,160 +1,543 @@
...
+        outputs:
+          - "success"
+        inputs:
+          - "peer"
+        version: "0.1.0"
+
+flows:
+  CreateVolume:
+    enabled: True
+    tendrl_context: "$cluster.id"
```
r0h4n: this needs more eyes, IMO this should be a list instead of single item

etc/sds_operations_gluster-3.8.3.yaml
```diff
@@ -1,160 +1,543 @@
...
+        help: "If force execute the action"
 
     atoms:
-        info:
-          type: "get"
         create:
-          run: "gluster_bridge.atoms.volume.create"
+          enabled: True
+          uuid: "242f6190-9b37-11e6-950d-a24fc0d9649c"
+          flows: ["CreateVolume"]
```
shtripat: Do you to have flows like CreateDistributeVolume, CreateStripedVolume, CreateStripedReplicateVolume etc as separate flows?
Based on non-mandatory fields possible for gluster volumes this permutation is going to be lot big in no I feel. Dont you feel we should some mechanism to mention optional arguments and only if they are passed we should invoke flow with different arguments?

@brainfunked @nthomas-redhat @nnDarshan @anmolbabu thoughts?
I strongly feel we should have some mechanism to mention a optional list of arguments as well for flows and this is the right juncture when we are defining the structure of yaml.

etc/sds_operations_gluster-3.8.3.yaml
```diff
@@ -1,160 +1,543 @@
...
-
-          global_inputs:
-          - An immutable input_name => value dictionary which specifies any initial inputs that should be automatically injected into the atoms scope before the atom execution commences (this allows for providing atom local values that do not need to be provided by other atoms/dependents).
-          local_inputs:
-          - A set or list of required inputs for this atom’s execute method
-          version: An immutable version that associates version information with this atom. It can be useful in resuming older versions of atoms. Standard major, minor versioning concepts should apply
+            - "volname"
+          inputs:
+            - "volname"
+            - ["stripe_count"]
```
shtripat: I didnt have a way to mention an optional argument so used square brackets (I feel it cannot be as square bracket means list here).
Now again my question is to have some way to mention optional arguments for the flows else we might end writing so many flows just because of optional arguments and their combinations in gluster.
I remember VDSM doing the same with optional arguments and we are doing something similar here.

shtripat: @r0h4n I have another question regarding listing of fields under `inputs` and `outputs`. Do we just need listing as its done in this patch or we need something like 

input: ['volname', 'brickdetails', 'force']

Kindly confirm with one example for `inputs` and `outputs` here may be ..

etc/sds_operations_gluster-3.8.3.yaml
```diff
@@ -1,160 +1,543 @@
...
+          enabled: True
+          uuid: "242f6190-9b37-11e6-950d-a24fc0d9649c"
+          flows: ["CreateVolume"]
+          run: "tendrl.gluster_bridge.atoms.volume.create.Create"
           type: "create"
-          conditions:
-            pre_condition : ["atoms.<atom_name>.<condition_function>1"...]
-            post_condition: ["atoms.<atom_name>.<condition_function>1"...]
-          name: Meaningful name for this atom, should be something that is distinguishable and understandable for notification, debugging, storing and any other similar purposes
+          pre_run : []
```
r0h4n: Yes, and these functions will receive the "inputs" for that atom for validation of those "inputs"

etc/sds_operations_gluster-3.8.3.yaml
```diff
@@ -1,160 +1,543 @@
...
+          post_run: []
```
r0h4n: Yes, we can have things done in post_run, but these actions should be atomic (reversible by the same atom)

etc/sds_operations_gluster-3.8.3.yaml
r0h4n: Its just a list for "inputs" and for "output" you need to use the structure described here 
https://github.com/Tendrl/bridge_common/blob/master/etc/samples/tendrl_definitions_generic-0.1.sample.yaml#L58

etc/sds_operations_gluster-3.8.3.yaml
```diff
@@ -1,160 +1,543 @@
...
+          outputs:
+            - "success"
+          inputs:
+            - "volname"
+          version: "0.1.0"
   brick:
+    enabled: True
+    value: "/clusters/$cluster.id/volumes/$volume.id/bricks/$brick.id"
     attrs:
       brickdetails:
```
r0h4n: then it is ok

etc/sds_operations_gluster-3.8.3.yaml
shtripat: Do you mean all the input fields would be passed for all the validate functions? It doesnt seem too good to me, but we can live with it I feel.

etc/sds_operations_gluster-3.8.3.yaml
shtripat: I have sent another patch post these comments with few validations added. Can you please check once and see if looks better?

etc/sds_operations_gluster-3.8.3.yaml
shtripat: Is it something like

outputs: [
  volname: {required: True},
  brickdetails: {required: True},
  replica_count: {required: False}
]

inputs: [volname, brickdetails, replica_count, ...]

For this reason only I wanted a working sample here :) Anyway plz confirm the same.

shtripat:
@r0h4n @nthomas-redhat @brainfunked @anmolbabu @nnDarshan I have pushed fresh patch with latest input / output guidelines for atoms and flows with dummy validations added for volume/peer atoms.

Kindly review..

r0h4n:
Fix "tox -epep8" failures https://travis-ci.org/Tendrl/gluster_bridge/builds/171148366

shtripat:
@brainfunked @nthomas-redhat waiting for your reviews. please do the same.

r0h4n:
Is this PR still required @shtripat

r0h4n:
@shtripat Is this still required? Please attach spec or bug

shtripat:
> @shtripat Is this still required? Please attach spec or bug

@r0h4n I would close this as its pretty old one and lot of changes have happened down the line.

PR #35, opened 2016-10-25
Gluster_bridge: Testcases for gluster bridge

GowthamShanmugam:
I have also modified infinite while loop in rpc.py
    - I have added check for infinite while loop in rpc.py

Signed-off-by: GowthamShanmugam gshanmug@redhat.com


GowthamShanmugam:
@r0h4n , @shtripat , @nthomas-redhat  please review this PR

tendrl/gluster_bridge/manager/rpc.py
```diff
@@ -16,15 +16,16 @@
 
 class EtcdRPC(object):
 
-    def __init__(self):
+    def __init__(self, Etcdthread):
```
GowthamShanmugam: To stop infinite loop in rpc when shutdown method called

PR #57, opened 2016-11-15
Update servers.py to correct peers object name in list view

shtripat:


shtripat:
@r0h4n @nthomas-redhat please review

PR #63, opened 2016-11-18
Remove volume from central store while delete flow execution

shtripat:
tendrl-bug-id: https://github.com/Tendrl/gluster_integration/issues/73

### DIFF
--- a/tendrl/gluster_integration/objects/volume/atoms/delete.py
+++ b/tendrl/gluster_integration/objects/volume/atoms/delete.py
@@ -26,6 +26,6 @@ class Delete(BaseAtom):
             ]
         )
         etcd_client = parameters['etcd_client']
-        vol_key = "clusters/%s/Volumes/%s/deleted" % (cluster_id, vol_id)
-        etcd_client.write(vol_key, "True")
+        vol_key = "clusters/%s/Volumes/%s" % (cluster_id, vol_id)
+        etcd_client.delete(vol_key, recursive=True)
         return True
--- a/tendrl/gluster_integration/manager/manager.py
+++ b/tendrl/gluster_integration/manager/manager.py
@@ -76,7 +76,6 @@ class GlusterIntegrationManager(Manager):
         ).__init__(
             "sds",
             cluster_id,
-            None,
             config,
             GlusterIntegrationSyncStateThread(self, cluster_id),
             GlusterIntegrationEtcdPersister(config),
--- a/tendrl/gluster_integration/tests/test_delete_volume_atom.py
+++ b/tendrl/gluster_integration/tests/test_delete_volume_atom.py
@@ -1,0 +1,43 @@
+import etcd
+import subprocess
+
+from mock import MagicMock
+from tendrl.gluster_integration.objects.volume.atoms.delete import Delete
+
+
+class TestDeleteVolume(object):
+    def test_run(self):
+        subprocess.call = MagicMock(
+            return_value='done'
+        )
+
+        parameters = {}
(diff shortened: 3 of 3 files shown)


shtripat:
@r0h4n @brainfunked @nnDarshan @nthomas-redhat please review.

tendrl/gluster_integration/objects/volume/atoms/delete.py
```diff
@@ -24,6 +24,6 @@ def run(self, parameters):
             ]
         )
         etcd_client = parameters['etcd_client']
-        vol_key = "clusters/%s/Volumes/%s/deleted" % (cluster_id, vol_id)
-        etcd_client.write(vol_key, "True")
+        vol_key = "clusters/%s/Volumes/%s" % (cluster_id, vol_id)
+        etcd_client.delete(vol_key, recursive=True)
```
r0h4n: are we not just marking the vol as deleted?

r0h4n:
This needs a github issue and unit tests

tendrl/gluster_integration/objects/volume/atoms/delete.py
shtripat: Marking the volume as deleted = True was a stop gap arrangement. Final deletion of the volume entry from etcd is done here.
Will update the test case for this.

r0h4n:
This needs a spec and/or issue correctly linked in commit msg

shtripat:
@r0h4n Isnt it just a bug fix? Do we really need a spec for this small change as well (overkill I feel) ?? There is a issue already and I would link in commit message.

shtripat:
> should we remove the "deleted" attr under volume from the definition file ?

@nnDarshan we might want to retain the field for use while soft delete (if we support de-comission for cluster later).

@r0h4n @brainfunked @nthomas-redhat thoughts??

shtripat:
> We must retain the object data and use "deleted" flag and filter it in tendrl-api for list views

@r0h4n Still I am not able to understand the intention behind retaining the object details for an object which is no more existing in underlying cluster. Also filtering in API layet based on this flag is good, but if say a volume with name "vol1" git deleted and we still have the record in etcd. Now user can actually create a volume with same name again.

I can brutforce and write code which keeps creating and deleting the same named volume again and again. This way I can hog the etcd with lot of data soon. There should be a good intention behind retaining the object details. I feel its good for retaining events and alerts kind of data for future audit. Specifically for this kind of data some sort of purging policy is defined and they are moved out of active DB.

@brainfunked @nthomas-redhat @nnDarshan @anmolbabu comment??

r0h4n:
All I am saying is,for future data analytics, the object lifecycle (create, update, delete) should be captured either via storing the actual object forever or 

Storing alerts/notifications related to above forever  

@brainfunked comments? 

@Tendrl/tendrl_api @Tendrl/tendrl-core comments?

shtripat:
@r0h4n fixed the travis failures now.

nthomas-redhat:
> All I am saying is,for future data analytics, any Tendrl/sds object's lifecycle (create, update, delete) should be captured either by storing (in etcd) the actual object forever or
> 
> Storing (in etcd) alerts/notifications related to above object's lifecycle forever
> 
> @brainfunked comments?
> 
> @Tendrl/tendrl_api @Tendrl/tendrl-core comments?

@r0h4n , Keeping the deleted data in store for future analysis is always good. But my concern here is how long? At some point this will become unmanageable and might affect the performance as well. My suggestion would be to remove it from etcd and if required for future reference, archive it and keep it outside central store.

shtripat:
Its taken care in later PRs. Closing

PR #66, opened 2016-11-23
Update tendrl-glusterd.service

shtripat:
With PrivateTmp set as true, a new /tmp is created and isolated from the other one. All the data saved inside are deleted once the service is stopped.

This was causing read failure while reading data from /tmp/glusterd-state in gluster-inetgration service while is written by glusterd command.

### DIFF
--- a/tendrl-glusterd.service
+++ b/tendrl-glusterd.service
@@ -7,7 +7,7 @@ ExecStart=/usr/bin/tendrl-gluster-integration
 ExecReload=/bin/kill -HUP $MAINPID
 KillMode=process
 Restart=on-failure
-PrivateTmp=true
+PrivateTmp=false
 
 [Install]
 WantedBy=multi-user.target


shtripat:
@nthomas-redhat @TimothyAsir @brainfunked please review.

nthomas-redhat:
This needs to be dealt in a different way later. We shouldn't puting anything into /tmp and reading from there

brainfunked:
Don't know why this was merged, but I agree with @nthomas-redhat. We should definitely not be writing any data to `/tmp` that we depend upon. Anything written to `/tmp` is supposed to be completely throwaway. If there's any data that we need to persist across service restarts, it should be written under `/var/lib/tendrl`.

shtripat:
@brainfunked I have sent another PR to change the dir to /var/run for writting glusterd-state. I feel /var/run is the proper dir for writing any runtime data. Comments?

brainfunked:
`/var/run` is for storing pid files and such. `/var/lib/tendrl` is where this data needs to reside.

shtripat:
@brainfunked ok. We have done as /var/run for the moment and while refactor lets take care of all the such guidelines and good practices this time .

PR #67, opened 2016-11-23
Update manager.py

shtripat:
Each systemctl service gets its own /tmp variation and owns that. If we use /tmp for creating glusterd-state file, tendrl-glusterd service is not able to read from the same as the file is written by glusterd service.

Changed the path to write the file to /var/run rather and this solves the issue.


shtripat:
@nthomas-redhat @nnDarshan please review.

shtripat:
[https://github.com/Tendrl/gluster_integration/issues/68](url)

r0h4n:
Thanks, please add test case only for your change

PR #77, opened 2016-12-06
Gluster-integration: fix the log config path.

nnDarshan:
Fixed the logging config path of common in gluster_integration conf file.

tendrl-bug-id: Tendrl/gluster_integration#64


nnDarshan:
@r0h4n @shtripat Please review

PR #79, opened 2016-12-08
Added logic to populate volume options to central store

shtripat:
As per new output from gluster get-state command, volume
option details are listed. Modiifed to parse and populate
the volume options details to the central store.

tendrl-bug-id: gluster_integration/#41
tendrl-spec: refactor_gluster_integration_get_state_dump.adoc

### DIFF
--- a/tendrl/gluster_integration/manager/manager.py
+++ b/tendrl/gluster_integration/manager/manager.py
@@ -237,7 +237,7 @@ class GlusterIntegrationManager(Manager):
                 except KeyError:
                     break
             # poplate the volume options
-            reg_ex = re.compile("^volume[0-9].options+")
+            reg_ex = re.compile("^volume[0-9]+.options+")
             options = {}
             for key in volumes.keys():
                 if reg_ex.match(key):
--- a/tendrl/gluster_integration/manager/manager.py
+++ b/tendrl/gluster_integration/manager/manager.py
@@ -1,6 +1,7 @@
 import gevent.event
 import json
 import logging
+import re
 import signal
 import subprocess
 import sys
@@ -20,6 +21,8 @@ from tendrl.gluster_integration.persistence.persister \
 from tendrl.gluster_integration.persistence.servers import Brick
 from tendrl.gluster_integration.persistence.servers import Peer
 from tendrl.gluster_integration.persistence.servers import Volume
+from tendrl.gluster_integration.persistence.servers import VolumeOptions
+
 from tendrl.gluster_integration.persistence.tendrl_context import TendrlContext
 from tendrl.gluster_integration.persistence.tendrl_definitions import \
     TendrlDefinitions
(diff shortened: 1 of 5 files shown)


shtripat:
@r0h4n @nthomas-redhat @brainfunked @anmolbabu @nnDarshan @GowthamShanmugam please review

shtripat:
Verified with latest nightly build of gluster from http://artifacts.ci.centos.org/gluster/nightly/release-3.9/7/x86_64/ and it works fine as expected.

tendrl/gluster_integration/manager/manager.py
```diff
@@ -200,6 +266,27 @@ def on_pull(self, raw_data, cluster_id):
                     index += 1
                 except KeyError:
                     break
+            # poplate the volume options
+            reg_ex = re.compile("^volume[0-9].options+")
```
nnDarshan: I think this doesn't match "volume12.options". Should it be something like ("^volume[0-9]+.options+")

shtripat: Ah. Thanks for pointing out. Will correct it.

tendrl/gluster_integration/manager/manager.py
```diff
@@ -200,6 +266,27 @@ def on_pull(self, raw_data, cluster_id):
                     index += 1
                 except KeyError:
                     break
+            # poplate the volume options
```
r0h4n: typo "poplate"

shtripat: Ah. done. will change

tendrl/gluster_integration/manager/manager.py
```diff
@@ -200,6 +266,27 @@ def on_pull(self, raw_data, cluster_id):
                     index += 1
                 except KeyError:
                     break
```
nthomas-redhat: Are all the keys(using in the above code) available all the time?
Should we really break out or continue?
Need logs here

tendrl/gluster_integration/manager/manager.py
```diff
@@ -200,6 +266,27 @@ def on_pull(self, raw_data, cluster_id):
...
+                    options[key] = volumes[key]
+            for key in options.keys():
+                volname = key.split('.')[0]
+                vol_id = volumes['%s.id' % volname]
+                dict = {}
+                for k, v in options.items():
+                    if k.startswith('%s.options' % volname):
+                        dict['.'.join(k.split(".")[2:])] = v
+                        options.pop(k, None)
+                self.persister.update_volume_options(
```
nthomas-redhat: does this function already implemented?

tendrl/gluster_integration/manager/manager.py
```diff
@@ -200,6 +266,27 @@ def on_pull(self, raw_data, cluster_id):
                     index += 1
                 except KeyError:
                     break
+            # poplate the volume options
+            reg_ex = re.compile("^volume[0-9]+.options+")
```
nthomas-redhat: I would really like to see this implemented as part of the main loop abd get this updated along with update_volume. The advantage we get is that we save 'n' number of extra updates to etcd. Is it something doable or you see any issues here?

tendrl/gluster_integration/manager/manager.py
```diff
@@ -149,15 +151,79 @@ def on_pull(self, raw_data, cluster_id):
                     self.persister.update_volume(
```
shtripat: I feel if we merge this first and then rebase refactoring patch on top would be easier as its just additional code. But anything is fine.

tendrl/gluster_integration/manager/manager.py
shtripat: So even if we move inside main loop of volume attributes, there also we would have to do at least the way its done for bricks and we wont be able to save no of writes to etcd.

tendrl/gluster_integration/manager/manager.py
shtripat: Yes the class is added for this as part of this PR only

tendrl/gluster_integration/manager/manager.py
nthomas-redhat: Make sense, but question here should we really need to break out if there is a KeyError raised while processing one of the volumes. I think we should not rather we should proceed further with other volumes. Does my understanding correct. Also we need to log the details as well here

tendrl/gluster_integration/manager/manager.py
nthomas-redhat: Well, my thought was to process and update the volume options along with other volume attributes and save additional updates to etcd. If you are convinced that this is the right way to do this, I am fine with it

tendrl/gluster_integration/manager/manager.py
nthomas-redhat: okay

tendrl/gluster_integration/manager/manager.py
shtripat: Thats a good point. I kept the code as it was earlier. We can think of logging the error for the failed volume and continue to next one.

shtripat: @nthomas-redhat the point is in valid case as well when all the volumes exhaust we depend on this KeyError to break the infinite while loop. If we change it to continue, it would never break. As we dont know the actula no of volume beforehand, I am not sure we can use a for loop to achieve this. We might have to retain the logic as is I feel. Chances of KeyError are not possible as data comes from a gluster generated file, so it not a risky deal I feel :)

nthomas-redhat: As discussed, we need to create a well formed output file so that we can handle the error conditions appropriately. Can do it as a separate patch, please raise an issue and attach here

shtripat: Created https://github.com/Tendrl/gluster_integration/issues/99 to track this change in a separate PR.

PR #96, opened 2016-12-29
Introduce flows specific to objects.

nnDarshan:
This patch segregates flows specific to an object
under the "flows" of that object instead of the
global namespace.

tendrl-spec: segragate_object_specific_flows
tendrl-bug-id: Tendrl/gluster_integration#95


nnDarshan:
@shtripat @nthomas-redhat Please review

nnDarshan:
@shtripat I feel there is no need for any code change for this needed in back end right? code change is needed only in API layer right ?

shtripat:
@nnDarshan thats correct. @anivargi can you please comment on this?

tendrl/gluster_integration/manager/tendrl_definitions_gluster.py
```diff
@@ -192,7 +143,56 @@
           run: tendrl.gluster_integration.objects.Volume.atoms.volume_stopped.VolumeStopped
           type: Check
           uuid: 242f6190-9b37-11e6-950d-a24fc0d9656c
-      attrs:
+      flows:
+        DeleteVolume:
+          atoms:
```
shtripat: As per latest refactoring `description` should be renamed as `help`

nnDarshan: Done

PR #97, opened 2016-12-30
Read cluster-id from a file instead of argument

nnDarshan:
Changed gluster_integration service to read cluster_id
from a file /var/lib/tendrl/gluster_cluster_id instead
of passing cluster_id to process as an argument.

tendrl-bug-id:Tendrl/gluster_integration#85


nnDarshan:
@nthomas-redhat @shtripat @r0h4n Please review

tendrl/gluster_integration/manager/utils.py
```diff
@@ -4,24 +4,18 @@
 import subprocess
 
 LOG = logging.getLogger(__name__)
-TENDRL_CONTEXT = "/etc/tendrl/gluster_integration/tendrl_context"
+TENDRL_CONTEXT = "/var/lib/tendrl/gluster_cluster_id"
```
r0h4n: the reason we were using "tendrl_context" was that it is a pre defined object in Tendrl

the new path "/var/lib/tendrl/gluster_cluster_id", how did you come up with this?

nnDarshan: The reason I changed the path was, since its in etc directory it might be considered as a config file by the user and users might meddle with the content. Since this is file is to be written and read only by tendrl(and this need not be exposed to user) felt /var/lib would be right path.

Please provide your thoughts.

shtripat: In that case shouldnt this be under /var/run/tendrl so something like that?

nnDarshan: @r0h4n @nthomas-redhat is  /var/run/tendrl fine with you guys ?

nnDarshan: reverted to older path itself

PR #102, opened 2017-01-05
Populate the current node's id as member in manager

shtripat:
tendrl-bug-id: Tendrl/gluster_integration#/101
tendrl-spec: multiple_node_ids_in_jobs.adoc


shtripat:
I would make node_id an optional argument in common module and gluster/cep integrations need not pass this.

shtripat:
Not required

PR #111, opened 2017-01-16
Re-factored to use load_config from commons

shtripat:


shtripat:
@Tendrl/tendrl-core please review

tendrl/gluster_integration/manager/manager.py
```diff
@@ -69,16 +73,20 @@ def _run(self):
 class GlusterIntegrationManager(Manager):
     def __init__(self, cluster_id):
         self._complete = gevent.event.Event()
+        etcd_kwargs = {
+            'port': config['etcd_port'],
```
anmolbabu: We probably need to discuss and close out on whether we are going with the approach here or config['configuration']['etcd_port'] as in https://github.com/Tendrl/node-agent/pull/143/files as I am also following that approach in my monitoring PR s.

shtripat:
Taken care as part of https://github.com/Tendrl/gluster-integration/pull/117

PR #117, opened 2017-01-22
Object tree

shtripat:


TimothyAsir:
Which rpm package provides namespaces module.
I could not able to see any python-namespaces rpm in epel

tendrl/gluster_integration/central_store/etcd_utils.py
```diff
@@ -0,0 +1,16 @@
+from tendrl.commons.etcdobj import fields
```
r0h4n: remove this file

tendrl/gluster_integration/objects/base_object.py
```diff
@@ -0,0 +1,96 @@
+import abc
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class BaseObject(object):
```
r0h4n: remove this, use commons baseobject

Issue #109, opened 2017-01-13
Fix package version issue


TimothyAsir:
Package version should comes from tendrl/gluster_integration/__init__.py and this should automatically updated from the github if we run make gitversion.

TimothyAsir:
Its fixed now

Issue #108, opened 2017-01-12
[rpm packaging] Dependency on tendrl-common


fbalak:
`tendrl-gluster-integration-0.0.1-1.el7.centos.noarch.rpm ` have a dependency on `tendrl-common`. `tendrl-common` was recently renamed to `tendrl-commons` so the package should reflect it in a dependency list.
### Currently:
```
# rpm -qp https://copr-be.cloud.fedoraproject.org/results/tendrl/tendrl/epel-7-x86_64/00498961-tendrl-gluster-integration/tendrl-gluster-integration-0.0.1-1.el7.centos.noarch.rpm --requires
/bin/sh
.
.
systemd
tendrl-common
rpmlib(PayloadIsXz) <= 5.2-1
#
```

### Should be:
```
# rpm -qp https://copr-be.cloud.fedoraproject.org/results/tendrl/tendrl/epel-7-x86_64/00498961-tendrl-gluster-integration/tendrl-gluster-integration-0.0.1-1.el7.centos.noarch.rpm --requires
/bin/sh
.
.
systemd
tendrl-commons
rpmlib(PayloadIsXz) <= 5.2-1
#
```

nthomas-redhat:
https://github.com/Tendrl/gluster-integration/pull/110

fbalak:
Tested with tendrl-gluster-integration-1.2-01_25_2017_18_29_53.noarch.rpm. This issue is resolved.

PR #124, opened 2017-01-25
Fix package version and tendrl-commons dependency issue

TimothyAsirJeyasing:
Currently we have to update the version in many places.
This patch provides an easy way to update the package version by
using the following machanism.

Updated Makefile to pickup package version from __init__.py file
and add a section called "gitversion" in Makefile to update
version in spec file based on github.
Updated setup.py file to pickup version from version.py file

tendrl-bug-id: Tendrl/gluster-integration#108
tendrl-bug-id: Tendrl/gluster-integration#109
Signed-off-by: Timothy Asir J <tjeyasin@redhat.com>


TimothyAsirJeyasing:
@ktdreyer , @r0h4n , @nthomas-redhat , @shtripat , Please review

PR #136, opened 2017-02-02
Update __init__.py

shtripat:
Corrected the name tendrl context variable


tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
```diff
@@ -10,7 +10,7 @@ def __init__(self, *args, **kwargs):
         super(Delete, self).__init__(*args, **kwargs)
 
     def run(self):
-        cluster_id = self.parameters['Tendrl_context.cluster_id']
+        cluster_id = self.parameters['TendrlContext.cluster_id']
```
r0h4n: cluster_id is called integration_id

shtripat: Done

Issue #143, opened 2017-02-22
Have a field for cluster name under tendrl-context.


nnDarshan:
We should have a field called integration name under TendrlContext object to indicate cluster name. Since
glusterfs backend does not have a concept of cluster name, its generated using integration_id as follows:
"glusterfs_<integration-id>" whaere integration_id is tendrl generated cluster id.

PR #144, opened 2017-02-22
Have a field for cluster name under tendrl-context.

nnDarshan:
We should have a field called integration name under
TendrlContext object to indicate cluster name. Since
glusterfs backend does not have a concept of cluster
name, its generated using integration_id as follows:
"glusterfs_<integration-id>" whaere integration_id is
tendrl generated cluster id.

tendrl-bug-id: Tendrl/gluster-integration#143


nnDarshan:
@r0h4n @nthomas-redhat @anivargi Please review

PR #152, opened 2017-02-26
Added extra check while gstatus run and percent use of cluster

shtripat:


shtripat:

PR #156, opened 2017-03-01
Save function for push job object in etcd

GowthamShanmugam:

### DIFF
--- a/tendrl/gluster_integration/central_store/__init__.py
+++ b/tendrl/gluster_integration/central_store/__init__.py
@@ -44,3 +44,6 @@ class GlusterIntegrationEtcdCentralStore(central_store.EtcdCentralStore):
 
     def save_utilization(self, utilization):
         tendrl_ns.etcd_orm.save(utilization)
+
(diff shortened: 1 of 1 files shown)


tendrl/gluster_integration/central_store/__init__.py
```diff
@@ -44,3 +44,6 @@ def save_nodecontext(self, node_context):
 
     def save_utilization(self, utilization):
         tendrl_ns.etcd_orm.save(utilization)
+
+    def save_job(self, job):
```
shtripat: Isnt this done as part another PR?

GowthamShanmugam: @shtripat yes job class in common right,  but save method should be in central_store to store that  job object, so i have created job class in common and save method in all component.

GowthamShanmugam: @shtripat if the component want to create any job, job is created using job class in common and stored via save method of its own (https://github.com/Tendrl/commons/pull/175)

Issue #167, opened 2017-03-13
Add gdeploy-wrapper in gluster integration


nnDarshan:
Add gdeploy-wrapper which acts as an interface between gluster integration and python-gdeploy. Atoms from
gluster integration will perform gluster related CRUD actions by invoking gdeploy-wrapper which in turn invokes python-gdeploy.

r0h4n:
Fixed

PR #186, opened 2017-03-17
Add gdeploy-wrapper for performing gluster actions

nnDarshan:
This patch adds gdeploy-wrapper which acts as an interface
between gluster integration and python-gdeploy. Atoms from
gluster integration will perform gluster related CRUD actions
by invoking gdeploy-wrapper which in turn invokes python-
gdeploy

tendrl-bug-id: Tendrl/gluster-integration#167


nnDarshan:
@r0h4n @nthomas-redhat @shtripat Please review

r0h4n:
All our gdeploy wrapper code should go into commons (some parts are used by node-agent and some by gluster-integration), let me know what you think

nnDarshan:
@r0h4n The interfaces that the wrapper provides are very different in each case. In case of gdeploy-wrapper in node-agent it provides interfaces for provisioning related actions like package installation, node_setup for gluster, cluster creation and disk provisioning.
However in case of wrapper in gluster-integration the wrapper provides interface for gluster specific actions for gluster volume management.
So, I feel they fine as is. 
@shtripat Please provide your thoughts

shtripat:
Agree @nnDarshan. I feel we should keep as is.

Issue #190, opened 2017-03-20
Add flows to create gluster bricks


nnDarshan:

r0h4n:
Please update

nnDarshan:
https://github.com/Tendrl/gluster-integration/pull/191 takes care of this

PR #191, opened 2017-03-20
Create brick

nnDarshan:


nnDarshan:
@shtripat @nthomas-redhat @r0h4n Please review

tendrl/gluster_integration/objects/definition/gluster.py
```diff
@@ -26,6 +26,22 @@
       type: Create
```
r0h4n: We are using yaml files to store definitions now, please remove this file

nnDarshan:
WIP- NOT verified yet

etc/create_brick.py
```diff
@@ -0,0 +1,37 @@
...
+job_id = str(uuid.uuid4())
+
+payload = {
+    "integration_id": integration_id,
+    "run": "tendrl.gluster.flows.CreateGlusterBrick",
+    "parameters": {
+        "GlusterBrick.brick_dict": {
+            "10.70.43.11:/mnt/tendrl_sample4/b7": {
+                "node_id": "4a47db22-b286-44da-8240-c8256b28d256",
+                "disk": "vdh"
```
shtripat: Dont we say `/dev/vdh`

etc/create_brick.py
```diff
@@ -0,0 +1,37 @@
...
+                "disk": "vdi"
+            }
+        }
+    },
+    "type": "sds",
+    "node_ids": ["4a47db22-b286-44da-8240-c8256b28d256"]
+}
+
+def run():
+    client = etcd.Client(host="10.70.43.135", port=2379)
```
shtripat: remove the actual IP

etc/create_brick.py
nnDarshan: gdeploy will figure that out.

etc/create_brick.py
nnDarshan: ack

etc/create_brick.py
```diff
@@ -0,0 +1,36 @@
...
+
+integration_id = "ab3b125e-4769-4071-a349-e82b380c11f4"
+job_id = str(uuid.uuid4())
+
+payload = {
+    "integration_id": integration_id,
+    "run": "tendrl.gluster.flows.CreateGlusterBrick",
+    "parameters": {
+        "GlusterBrick.brick_dict": {
+            "12.34.56.78:/mnt/tendrl_sample4/b7": {
```
brainfunked: It must be possible to supply tuning configuration per brick, such as whether the brick is backed by a RAID device or just a regular disk etc.

tendrl/gluster_integration/manager/__init__.py
```diff
@@ -71,6 +71,9 @@ def main():
     pm = ProvisioningManager("GdeployPlugin")
     NS.gdeploy_plugin = pm.get_plugin()
 
+    pm = ProvisioningManager("GdeployPlugin")
```
nnDarshan: Done

etc/create_brick.py
```diff
@@ -0,0 +1,36 @@
...
+                "node_id": "4a47db22-b286-44da-8240-c8256b28d256",
+                "disk": "vdh"
+            },
+            "12.34.56.78:/mnt/tendrl_sample5/b8": {
```
brainfunked: Given that we're allowing bricks to be created without being attached to a volume, it'll be necessary to standardise a path under which tendrl will always mount the bricks and expect bricks to be mounted, if created manually, to be able to detect them. In such a scenario, we cannot ask for a brick path, but just a name. This also ties in to the comment below against the `Create.run` method.

However, the mount path will not be available to tendrl when an existing gluster is imported. In such a case, a mount path could specified as a parameter per brick and tendrl would need to index all the mount paths, including it's own standard path, and scan all of them for syncing available bricks.

tendrl/gluster_integration/objects/gluster_brick/atoms/create/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+                brick_dict[host] = {}
+            disk = v["disk"]
+            brick_path = k.split(":")[-1]
+            mount_path = "/" + "/".join(brick_path.split("/")[1:-1])
+            brick_dict[host].update({
+                disk: {
+                    "node_id": v["node_id"],
+                    "mount_path": mount_path,
+                    "brick_path": brick_path,
+                    "lv": "tendrl" + brick_path.replace("/", "_") + "_lv",
```
brainfunked: Here, tendrl is making several conventions regarding naming. As discussed above, such conventions need to be applied to standardise a mount path where tendrl would be able to pick up the available bricks from. Further implication is that the sync operation needs to scan the mount path and gather the mapping of available bricks their corresponding devices.

etc/create_brick.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+integration_id = "ab3b125e-4769-4071-a349-e82b380c11f4"
+job_id = str(uuid.uuid4())
+
+payload = {
+    "integration_id": integration_id,
+    "run": "tendrl.gluster.flows.CreateGlusterBrick",
+    "parameters": {
+        "GlusterBrick.brick_dict": {
```
nnDarshan: Done

etc/create_brick.py
nnDarshan: Gdeploy currently has this limitation, We cant specify different configuration per brick. It has to be same for all the bricks passed.

etc/create_brick.py
nnDarshan: Done, have standardized the brick path to "/glusterfs"  user now has to provide only brick name. the brick will be mounted under "/glusterfs"

tendrl/gluster_integration/objects/gluster_brick/atoms/create/__init__.py
nnDarshan: Sure. Have raised an issue: https://github.com/Tendrl/gluster-integration/issues/221
Will take this up soon

tendrl/gluster_integration/objects/definition/gluster.yaml
```diff
@@ -1,4 +1,5 @@
 namespace.gluster:
+  gluster_brick_path: "/glusterfs"
```
nnDarshan: Your point makes sense.
I just had a discussion with shubhendu and RHSC folks(old gluster console)  and got inputs that it would be better to make it configurable. So am planning to move this ti gluster_integration conf file and have "tendrl_gluster" as default
@r0h4n what do you feel ?

r0h4n: Makes sense, go ahead, cc'ing @brainfunked

nnDarshan:
https://github.com/Tendrl/gluster-integration/issues/190

PR #197, opened 2017-03-22
Corrected an issue while starting gluster-integration

shtripat:
While starting and syncing the cluster details, the namespace
variables were not initialized properly. So synced cluster data
used to be wrong.

tendrl-bug-id: Tendrl/gluster-integration#198


tendrl/gluster_integration/manager/__init__.py
```diff
@@ -34,10 +34,12 @@ def main():
     NS.state_sync_thread = sds_sync.GlusterIntegrationSdsSyncStateThread()
 
     NS.node_context.save()
-    NS.tendrl_context.save()
-    NS.gluster_integration.definitions.save()
-    NS.gluster_integration.config.save()
+    NS.gluster.tendrl_context.save()
```
r0h4n: tendrl_context is global object

let it be NS.tendrl_context

tendrl/gluster_integration/manager/__init__.py
```diff
@@ -34,10 +34,12 @@ def main():
...
+    NS.gluster.definitions.save()
+    NS.gluster.config.save()
     NS.publisher_id = "gluster_integration"
+    NS.tendrl_context = \
```
r0h4n: this is not required, already done in https://github.com/Tendrl/commons/blob/develop/tendrl/commons/__init__.py#L70

tendrl/gluster_integration/objects/node_context/__init__.py
```diff
@@ -37,7 +37,7 @@ def _create_node_id(self, node_id=None):
         with open(local_node_context, 'wb+') as f:
```
r0h4n: This object is moved to comons, please delete from here

tendrl/gluster_integration/objects/tendrl_context/__init__.py
```diff
@@ -26,7 +26,7 @@ def create_local_integration_id(self):
         with open(tendrl_context_path, 'wb+') as f:
```
r0h4n: This object is moved to comons, please delete from here

tendrl/gluster_integration/manager/__init__.py
shtripat: `NS.tendrl_context` doesnt contain `integration_id` populated if just `tendrl-gluster-integration` started.
The field `NS.tendrl_context.integration_id` is populated only if import/create cluster flow is invoked. By any chance if services go down and re-started, there is no way integration_id would be populated. This would cause issues while syncing the cluster details and would add details to etc in the form `clusters//Volumes`. This would end up data in wrong state.

tendrl/gluster_integration/manager/__init__.py
shtripat: In accordance to above comment, if cluster specific `tendrl_context` is loaded, same should be available under NS.tendrl_context, so that all the places we refer `NS.tendrl_context.integration_id` works seamlessly.

tendrl/gluster_integration/objects/node_context/__init__.py
shtripat: If there is no way to load `tendrl_context` from local, there is no way we get the details after re-start of `tendrl-gluster-integration` and as mentioned above all the cluster sync details would be wrong.

tendrl/gluster_integration/objects/tendrl_context/__init__.py
shtripat: If there is no way to load `tendrl_context` from local, there is no way we get the details after re-start of `tendrl-gluster-integration` and as mentioned above all the cluster sync details would be wrong.
So effectively we still need loading of `tendrl_context` from local for re-start scenario.
@r0h4n if you any other suggestion, plz provide details here.

tendrl/gluster_integration/objects/node_context/__init__.py
shtripat: Don't we plan to maintain node context specific to cluster anymore? At the moment I am aware of consequences of removing this.

tendrl/gluster_integration/manager/__init__.py
shtripat: I tried using `NS.tendrl_context.save()` only in manager now, but the issue is from this `save()` function when it tries to load `tendrl_context`, its an empty object and doesn't contain `integration_id` populated. Thus the whole issue crops up again back and cluster details are written like `clusters//Volumes` etc to central store.
Due to this I feel in `gluster-integration` we do need logic to load `tendrl_context` from local which should be written while import/create cluster flow. Otherwise from `gluster-integration` when we initialize `NS.tendrl_context`, it should rather read details from etcd and assign to NS somehow.
@r0h4n comments?

shtripat: This problem happens even with `node-agent` if re-started and no create/import executed after that. Even if earlier we had ran create/import and then the `integration_id` was populated. But after re-start again `tendrl_context.integration_id` goes blank.

tendrl/gluster_integration/objects/node_context/__init__.py
r0h4n: A cluster will have multiple nodes, which context are you going to save to /clusters? We are already saving ClusterNodeContext which is done by the node-agent.

Please remove this object

tendrl/gluster_integration/manager/__init__.py
r0h4n: 1) TendrlContext is save to etcd by Import/Create flows
2) To load() TendrlContext from etcd you need only NS.node_context.node_id

3) Put this fix for ceph-integration, gluster-integration , performance_mon, alerting (cc: @anmolbabu )
https://github.com/Tendrl/node-agent/commit/15dee0f2280a8290a6d25f4a8d5425ef79750b43

4) Above fix basically loads the TendrlContext if any (from etcd) at Tendrl service startup 

This way we dont need a disk local TendrlContext 

Let me know we need to debug this more

anmolbabu: No tendrl_context required in alerting and performance_monitoring applications currently(after integration id is made optional)....

shtripat: I agree that using `NS.node_context.node_id` we can load `tendrl_context` and get the `integration_id` from the same.
But the issue here is that after re-start even the field `nodes/{node-id}/TendrlContext/integration_id` goes blank if no create/import executed after re-start. So effectively even if we use `NS.node_context.node_id` to load the `tendrl_context` its not going to help.
@r0h4n here we need to fix the node-agent re-start and it should populate the previously created `tendrl_context` in NS

shtripat:
@r0h4n with latest rpm, the above changes work fine. Verified with import -> re-start of `node-agent` and `gluster-integration` and it works fine.

PR #220, opened 2017-04-13
Corrected check for non-existence of volume after delete

shtripat:
Also added enviornment variable `ANSIBLE_HOST_KEY_CHECKING=False`
for making sure first time gdeploy execution doesnt ask for
confirmation to connect to node.

### DIFF
--- a/tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
+++ b/tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
@@ -1,3 +1,4 @@
+import etcd
 import subprocess
 
 from tendrl.commons.event import Event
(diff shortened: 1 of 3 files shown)


shtripat:
@r0h4n @nnDarshan plz review

tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
```diff
@@ -83,13 +86,30 @@ def run(self):
...
+                ),
+                recursive=True
+            )
+        except etcd.EtcdKeyNotFound as ex:
+            Event(
+                ExceptionMessage(
+                    priority="debug",
+                    publisher=NS.publisher_id,
+                    payload={
+                        "message": "Given key is not present in etcd.",
```
r0h4n: what kinda msg is this? 

Id prefer  "Cannot delete volume %vol_id, volume does not exist "

tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
```diff
@@ -83,13 +86,30 @@ def run(self):
...
-            "clusters/%s/Volumes/%s" % (
-                NS.tendrl_context.integration_id,
-                self.parameters['Volume.vol_id']
-            ),
-            recursive=True
-        )
+        # Intentionally keeping sleep with high value and blocking
+        # one and not the gevent.sleep(). Make sure no sync etc
+        # happens for this time and then execute etcd delete for volume
+        time.sleep(30)
```
r0h4n: put a check in NS.gdeploy_plugin.delete_volume to block till volume is actually deleted in sds, 

time.sleep doesnt belong here

shtripat: Problem here is that the volume gets deleted from gluster cluster but its not deleted from etcd. The gdeploy plugin has returned a success after deletion and then the below etcd delete key was success.

If we used to put a pdb to debug it always used to work, but in normal flow it never used to delete the key and so post check to verify if the volume is deleted successfully used to fail.

So actually waiting inside gdeploy plugin is not the place this is required. Problem is while deletion from etcd.

This explicit time.sleep() helped so kept it like this.

tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
shtripat: This is copy + paste issue :) Will correct the message.

tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
shtripat: I would add a while loop and wait for the deletion to happen here may be.

Issue #217, opened 2017-04-08
Status is not coming for gluster cluster


gnehapk:
For ceph cluster, status is coming under "globalDetails". But gluster status is not there in the response.

 Response of gluster cluster in GetClusterList call -
"clusters":[
{
"cluster_id":"6bedd29e-40e7-4457-bdea-97de56187924",
"cluster_name":"gluster-fce90e0ed3b1c08aef534b1ef34b1e7fcac70ddec21dbbd250ecf906cfabeb42",
"integration_id":"6bedd29e-40e7-4457-bdea-97de56187924",
"sds_name":"glusterfs",
"sds_version":"3.9.1",
"volumes":{},
"nodes":{
"46f9d97f-0692-49a7-9bcf-8fd4b2065abb":{
"node_id":"46f9d97f-0692-49a7-9bcf-8fd4b2065abb",
"status":"UP",
"tags":"[\"tendrl/integration/gluster\", \"gluster/server\", \"tendrl/node\"]",
"fqdn":"dhcp43-162.lab.eng.blr.redhat.com",
"machine_id":"f8dc286ad2a04b6bacb667115cd1fa5f"
},
"55297b9f-c563-447f-ae38-17f64f672b8a":{
"fqdn":"dhcp43-164.lab.eng.blr.redhat.com",
"machine_id":"d790359bdbca4d8d9d47bb2556ef47ea",
"node_id":"55297b9f-c563-447f-ae38-17f64f672b8a",
"status":"UP",
"tags":"[\"tendrl/integration/gluster\", \"gluster/server\", \"tendrl/node\"]"
}
}
}]

@nthomas-redhat @brainfunked @r0h4n @nnDarshan @shtripat Please check.

r0h4n:
Closing this one, @gnehapk I remember this was fixed, please verify

PR #224, opened 2017-04-17
Added import for `os` as it was removed as a regression

shtripat:
Due to this cluster status and volume utlization details
were not being synced while regular sync.

tendrl-bug-id: Tendrl/gluster-integration#217


shtripat:

Issue #237, opened 2017-04-20
If no volume in cluster, the locking for `volume` in state sync hangs


shtripat:
The acquire of lock while state sync in `tendrl-gluster-integration` hangs if there are no volumes in the cluster.
Need a fix for the same.

r0h4n:
The locks work via etcd dir _locks/ . Is there a race condition to acquire this lock?

nnDarshan:
I feel the issue might be this:

* when gluster-integration process that has the volume lock currently is killed/shut down. It wont release the lock. Because of this no other gluster-integration will be able to acquire the lock or  Even if you restart the process, it wont able to get the lock. 

Solution should be to delete the locks in etcd.

shtripat:
Thats what even I suspected. But the abrupt shutdown case should be handled while start I feel and we should remove the lock if any is etcd in that case.

nnDarshan:
@shtripat @r0h4n @shtripat 
Have sent a fix for this issue: https://github.com/Tendrl/gluster-integration/pull/238
Please let me know what you guys think

PR #238, opened 2017-04-20
Set ttl for the volume lock that we acquire

nnDarshan:
The volume lock that is acquired, should have a ttl
So that, even if the service is ended abruptly. The
lock wont end up for ever, as there is no one to
release it.

tendrl-bug-id: Tendrl/gluster-integration#237

### DIFF
--- a/tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
+++ b/tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
@@ -88,8 +88,23 @@ class Delete(objects.BaseAtom):
             try:
                 # Acquire lock before deleting the volume from etcd
                 # We are blocking till we acquire the lock
+                # the lock will live for 60 sec after which it will be released.
                 lock = etcd.Lock(NS.etcd_orm.client, 'volume')
-                lock.acquire(blocking=True,lock_ttl=None)
+
+                while not lock.is_acquired:
+                    try:
+                        # with ttl set, lock will be blocked only for 60 sec
+                        # after which it will raise lock_expired exception.
+                        # if this is raised, we have to retry for lock
+                        lock.acquire(blocking=True,lock_ttl=60)
+                        if lock.is_acquired:
+                            # renewing lock as we are not sure, how long we
+                            # were blocked before the lock was given.
+                            # NOTE: blocked time also counts as ttl
+                            lock.acquire(lock_ttl=60)
+                    except etcd.EtcdLockExpired:
+                        continue
+                
                 NS.etcd_orm.client.delete(
                     "clusters/%s/Volumes/%s" % (
                         NS.tendrl_context.integration_id,
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -30,9 +30,16 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
         while not self._complete.is_set():
             try:
                 # Acquire lock before updating the volume details in etcd
-                # We are blocking till we acquire the lock
+                # We are blocking till we acquire the lock.
+                # the lock will live for 60 sec after which it will be released.
                 lock = etcd.Lock(NS.etcd_orm.client, 'volume')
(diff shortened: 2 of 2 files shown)


nnDarshan:
@shtripat @r0h4n @nthomas-redhat PLease review

tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
```diff
@@ -88,8 +88,9 @@ def run(self):
             try:
                 # Acquire lock before deleting the volume from etcd
                 # We are blocking till we acquire the lock
+                # the lock will live for 60 sec after which it will be released.
                 lock = etcd.Lock(NS.etcd_orm.client, 'volume')
-                lock.acquire(blocking=True,lock_ttl=None)
+                lock.acquire(blocking=True,lock_ttl=60)
```
shtripat: This is 60 seconds?

nnDarshan: yes

shtripat:
LGTM

tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
```diff
@@ -88,8 +88,25 @@ def run(self):
...
+                locked = False
                 lock = etcd.Lock(NS.etcd_orm.client, 'volume')
-                lock.acquire(blocking=True,lock_ttl=None)
+
+                while not locked:
+                    try:
+                        # with ttl set, lock will be blocked only for 60 sec
+                        # after which it will raise lock_expired exception.
+                        # if this is raised, we have to retry for lock
+                        lock.acquire(blocking=True,lock_ttl=60)
```
r0h4n: whats the reasoning behind choosing a 60 second ttl? high ttls will increase idling of the sync loop and any other volume operation

The lock should live approximately as long as the lock is required. In case of delete, a 10 second TTL would suffice

tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
```diff
@@ -88,8 +88,25 @@ def run(self):
             try:
                 # Acquire lock before deleting the volume from etcd
                 # We are blocking till we acquire the lock
+                # the lock will live for 60 sec after which it will be released.
+                locked = False
```
r0h4n: you can directly use while lock.is_acquired, why add another boolean flag?

nnDarshan: Good point. will do that.

tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
nnDarshan: We are releasing the lock as soon as we are done with the job. So the ttl of 60 sec doesn't come into picture in normal case. Only if there is abrupt stopping of process while the lock was acquired, this ttl kicks in and releases the lock.
So to be on the safer side have set ttl of 60 sec. Considering this will happen rarely, I feel it to be okay
@r0h4n Thoughts ?

tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py
```diff
@@ -88,8 +88,23 @@ def run(self):
...
                 lock = etcd.Lock(NS.etcd_orm.client, 'volume')
-                lock.acquire(blocking=True,lock_ttl=None)
+
+                while not lock.is_acquired:
+                    try:
+                        # with ttl set, lock will be blocked only for 60 sec
+                        # after which it will raise lock_expired exception.
+                        # if this is raised, we have to retry for lock
+                        lock.acquire(blocking=True,lock_ttl=60)
+                        if lock.is_acquired:
```
nthomas-redhat: Think of scenarion where lock is aquired and before refreshing it got expired. how does this code will behave that time? will it try to aquire again or come out of the while loop lock.acquire is true(or will it reset to false when Lock gets expired?)

r0h4n: The lock is refreshed during the acquire call itself, I dont think this would be a problem

PR #239, opened 2017-04-24
Added default status and utilization for cluster with no volumes

shtripat:
In case there are no volumes in the cluster, gstatus doesnt return
the utilization and status details for the underlying cluster.

Added logic to calculate the status based on peer's connected status
and utilization to default to zeros.

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -267,6 +267,41 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                                     rebal_files=volume.rebal_files,
                                     rebal_data=volume.rebal_data
                                 ).save()
+
+                else:
+                    # If gstatus does not return any status details in absence
+                    # of volumes, default set the status and utilization details
+                    # A sample output from gstatus if no volumes in cluster is as below
+                    #
+                    # >gstatus -o json
+                    # This cluster doesn't have any volumes/daemons running.
+                    # The output below shows the current nodes attached to this host.
+                    #
+                    # UUID					Hostname    	State
+                    # 6a48d2f7-3859-4542-86e0-0a8146588f31	{FQDN-1}	Connected
+                    # 7380e303-83b6-4728-918f-e99029bc1bce	{FQDN-2}	Connected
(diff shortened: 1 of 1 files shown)


shtripat:
@nthomas-redhat plz review

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -267,6 +267,27 @@ def _run(self):
                                     rebal_files=volume.rebal_files,
                                     rebal_data=volume.rebal_data
                                 ).save()
+                # If gstatus does not return any status details in absence
+                # of volumes, default set the status and utilization details
+                out_lines = stdout.split('\n')
```
r0h4n: please provide sample of stdout

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -267,6 +267,27 @@ def _run(self):
...
+                connected = True
+                for index in range(4, len(out_lines) - 2):
+                    if out_lines[index].split('\t')[2].strip() != 'Connected':
```
r0h4n: are you sure the stdout can always be processed this way?

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: Sure. Will add one sample here.

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: I have seen output of gstatus is like this only always if volumes are not present.

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -267,6 +267,40 @@ def _run(self):
                                     rebal_files=volume.rebal_files,
                                     rebal_data=volume.rebal_data
                                 ).save()
+
```
nthomas-redhat: Are you running this code always even if the volumes are present? In that case there is a possibility of overwriting the status provided by gstatus?

shtripat: Ah it should be in else part of gstatus command. Will update

Issue #236, opened 2017-04-19
etcd does not preserve the order of bricks under the volume in etcd


nnDarshan:
Gluster integration does not preserve the order of bricks while storing them under gluster volume in etcd. The order is very important as it gives information about the subvolumes (replica sets / etasure sub volume).

This has to be fixed. We should either maintain the order of bricks or have one more level under volumes for sub volumes(under this bricks will be stored).

nnDarshan:
@shtripat @nthomas-redhat @r0h4n Please provide your inputs

r0h4n:
Brick id's can be sequential keys in etcd instead of Brick uuid (uuid can be just an attribute of brick)

https://github.com/jplana/python-etcd/blob/0.4.5/src/etcd/client.py#L451

Search: "Atomically Creating In-Order Keys"
https://coreos.com/etcd/docs/latest/v2/api.html

r0h4n:
@nnDarshan closing this one

PR #240, opened 2017-04-24
Store the gluster volume bricks in order

nnDarshan:
The order of gluster volume bricks is important
because, we figure out the sub-volume details by
seeing the order of bricks. The bricks were not
ordered before, this patch makes sure we save the
bricks in order

tendrl-bug-id: Tendrl/gluster-integration#236

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -47,22 +47,22 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                 )
                 raw_data = ini2json.ini_to_dict('/var/run/glusterd-state')
                 subprocess.call(['rm', '-rf', '/var/run/glusterd-state'])
-                NS.sync_object = NS.gluster.objects.\
+                sync_object = NS.gluster.objects.\
                     SyncObject(data=json.dumps(raw_data))
-                NS.sync_object.save()
+                sync_object.save()
 
                 if "Peers" in raw_data:
                     index = 1
                     peers = raw_data["Peers"]
                     while True:
                         try:
-                            NS.peer = NS.gluster.\
+                            peer = NS.gluster.\
                                 objects.Peer(
                                     peer_uuid=peers['peer%s.uuid' % index],
                                     hostname=peers['peer%s.primary_hostname' % index],
                                     state=peers['peer%s.state' % index]
                                 )
-                            NS.peer.save()
+                            peer.save()
                             index += 1
                         except KeyError:
                             break
@@ -71,7 +71,7 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                     volumes = raw_data['Volumes']
                     while True:
                         try:
-                            NS.volume = NS.gluster.objects.Volume(
+                            volume = NS.gluster.objects.Volume(
                                 vol_id=volumes[
                                     'volume%s.id' % index
                                 ],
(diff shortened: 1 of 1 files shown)


nnDarshan:
@shtripat @r0h4n @nthomas-redhat Please have a look

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -143,6 +143,18 @@ def _run(self):
                                 ],
                             )
                             NS.volume.save()
+
+                            try:
```
shtripat: you mean every sync we should delete all the bricks and add them back? Why so?

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -168,7 +180,20 @@ def _run(self):
                                                 'volume%s.brick%s.mount_options' % (
                                                     index, b_index))
                                         )
-                                    NS.brick.save()
+                                    # Store the bricks sequentially, as the order of bricks
+                                    # matter while figuring out the sub-volumes
+                                    b = NS.brick.__dict__.copy()
+                                    b.pop('_etcd_cls')
+                                    b['name'] = NS.brick.path.replace("/", "_")
+                                    NS.etcd_orm.client.write(
```
shtripat: Can we use brick object's save() function here?

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -168,7 +180,20 @@ def _run(self):
...
+                                        "clusters/%s/Volumes/%s/Bricks/" % (
+                                            NS.tendrl_context.integration_id,
+                                            NS.volume.vol_id,
+                                        ),
+                                        json.dumps(b),
```
shtripat: How we make sure order is maintained here?

tendrl/gluster_integration/sds_sync/__init__.py
nnDarshan: Now we are writing the bricks with "append=true". So each brick will have sequence_no as key and this maintains the order of bricks. If we don't delete it every time it gets accumulated.

tendrl/gluster_integration/sds_sync/__init__.py
nnDarshan: save method doesn't have the "append" parameter mentioned above

tendrl/gluster_integration/sds_sync/__init__.py
nnDarshan: As mentioned above, we use the sequence no. to figure out the order

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: Oh somehow I overlooked I feel. Thats fine..

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: ok

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: ok

shtripat:
LGTM

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -168,7 +180,20 @@ def _run(self):
                                                 'volume%s.brick%s.mount_options' % (
                                                     index, b_index))
                                         )
-                                    NS.brick.save()
+                                    # Store the bricks sequentially, as the order of bricks
+                                    # matter while figuring out the sub-volumes
```
r0h4n: Whats the reason to overload NS with NS.brick or NS.volume here? 

If there is no important reason, Id prefer we dont overload NS in this file at all. Please use local variables for brick, volume, vol_opts etc which are assigned NS right now

nnDarshan: Ack. Will change them

nnDarshan:
@r0h4n Have incorporated your suggestions

r0h4n:
Verified?

nnDarshan:
@r0h4n Yes its verified

Issue #245, opened 2017-04-28
Exceptions in gluster-integration


anmolbabu:
{u'bb0846e5-f9fa-4299-b9db-fb8c438973c4': {u'publisher': u'gluster_integration', u'job_id': u'', u'timestamp': u'2017-04-28 17:36:08.758784+00:00', u'caller': {u'function': u'_run', u'line_no': u'360', u'filename': u'/usr/lib/python2.7/site-packages/tendrl/gluster_integration/sds_sync/__init__.py'}, u'updated_at': u'2017-04-28 17:36:09.642893+00:00', u'payload': {u'exception_type': u'IndexError', u'message': u'error', u'exception_traceback': u'[{u\'function\': u\'run\', u\'line\': 327, u\'file\': u\'/usr/lib64/python2.7/site-packages/gevent/greenlet.py\', u\'statement\': u\'result = self._run(*self.args, **self.kwargs)\'}, {u\'function\': u\'_run\', u\'line\': 338, u\'file\': u\'/usr/lib/python2.7/site-packages/tendrl/gluster_integration/sds_sync/__init__.py\', u\'statement\': u"if out_lines[index].split(\'\\\\t\')[2].strip() != \'Connected\':"}]', u'exception': u'!!python/object/apply:exceptions.IndexError [list index out of range]\n'}, u'priority': u'error', u'parent_id': u'', u'node_id': u'9167cdb6-9891-43e9-b46d-9a1a5d4013de', u'cluster_id': u'', u'flow_id': u'', u'message_id': u'bb0846e5-f9fa-4299-b9db-fb8c438973c4'}}

anmolbabu:
This is not very consistent and happens very rarely

shtripat:
@anmolbabu just to confirm, in case of failures did you have any volumes in your setup? Also how many nodes are there in the trusted pool?

anmolbabu:
@shtripat There were 3 nodes in the cluster/trusted pool but I am not very sure if I had any volumes I think I had 1 volume

shtripat:
@anmolbabu I have tried with single node cluster, 4 node cluster, with volume and without volume scenarios and never saw the issue. Anyway looks like some corner case where `gstatus` is not returning the output in required formats.

Recently have seen bunch of issues raised against `gstatus`. Could be one of them but not very sure.

anmolbabu:
[...]

PR #248, opened 2017-05-08
Added check for proper output of gstatus

shtripat:
Make sure default status update doesnt fail with exception
even if gstatus return bogus data.

tendrl-bug-id: Tendrl/gluster-integration#245

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -334,8 +334,10 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                     out_lines = stdout.split('\n')
                     connected = True
                     for index in range(4, len(out_lines) - 2):
-                        if out_lines[index].split('\t')[2].strip() != 'Connected':
-                            connected = connected and False
+                        node_status_det = out_lines[index].split('\t')
+                        if len(node_status_det) > 2:
+                            if node_status_det[2].strip() != 'Connected':
+                                connected = connected and False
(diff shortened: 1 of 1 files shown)


shtripat:

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -334,8 +334,10 @@ def _run(self):
                     out_lines = stdout.split('\n')
                     connected = True
                     for index in range(4, len(out_lines) - 2):
-                        if out_lines[index].split('\t')[2].strip() != 'Connected':
-                            connected = connected and False
+                        node_status_det = out_lines[index].split('\t')
+                        if len(node_status_det) > 2:
+                            if node_status_det[2].strip() != 'Connected':
+                                connected = connected and False
```
anmolbabu: can this be connected=False?
I feel connected = connected and False is equivalent of connected=False....
Please correct me if I am missing something

shtripat: I want to AND the previous value with `False` if current node is not `Connected`.
So at the end if any of the nodes is not `Connected` we will get a False value.

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -334,8 +334,10 @@ def _run(self):
                     out_lines = stdout.split('\n')
                     connected = True
                     for index in range(4, len(out_lines) - 2):
```
r0h4n: please ensure out_lines is not None

Issue #252, opened 2017-05-15
The gluster bricks directory should segregate used and free bricks


nnDarshan:
Gluster bricks section in etcd, located at "clusters/cluster-id/nodes/node-id/GlusterBricks/" currently lists all the bricks available on the node. There is no way to figure out if it is a used/free brick. So we should have sections under this to segregate used and free bricks.

We should have 3 sections under glusterBricks directory namely:
all : should contain all the bricks with complete info about the bricks
free: should contain all the free bricks with link to actual brick in "all" section
used:  should contain all the used bricks with link to actual brick in "all" section

PR #256, opened 2017-05-16
Classify bricks gluster bricks as used/free

nnDarshan:
This patch classifies the gluster bricks as used and
free based on if its used in a gluster volume. Also
this patch adds a cluster level object to hold the
default gluster brick path

tendrl-bug-id: https://github.com/Tendrl/gluster-integration/issues/252

### DIFF
--- a/tendrl/gluster_integration/objects/definition/gluster.yaml
+++ b/tendrl/gluster_integration/objects/definition/gluster.yaml
@@ -24,7 +24,9 @@ namespace.gluster:
       type: Create
       uuid: 1951e821-7aa9-4a91-8183-e73bc8275b8e
       version: 1
-    CreateGlusterBrick:
+    CreateBrick:
+      tags:
+       - "provisioner/$TendrlContext.integration_id"
       atoms:
         - tendrl.gluster.objects.GlusterBrick.atoms.Create
       help: "Create gluster bricks that can be used by volumes"
@@ -36,7 +38,7 @@ namespace.gluster:
           - GlusterBrick.disk_type
           - GlusterBrick.disk_count
           - GlusterBrick.stripe_size
-      run: tendrl.gluster.flows.CreateGlusterBrick
+      run: tendrl.gluster.flows.CreateBrick
       type: Create
       uuid: 1951e821-7aa9-4a91-8183-e73bc8275b8d
       version: 1
@@ -59,6 +61,14 @@ namespace.gluster:
       list: clusters/$TendrlContext.integration_id/GlobalDetails
       value: clusters/$TendrlContext.integration_id/GlobalDetails
       help: Clustr global details
+    GlusterBrickDir:
+      attrs:
+        default_brick_dir:
+          help: gives the default brick directory
+          type: String
+      enabled: true
+      value: clusters/$TendrlContext.integration_id/GlusterBrickDir
+      help: directory of gluster bricks
     Utilization:
       attrs:
         raw_capacity:
@@ -166,7 +176,7 @@ namespace.gluster:
               - GlusterBrick.disk_type
               - GlusterBrick.disk_count
               - GlusterBrick.stripe_size
-          name: create_gluster_brick
+          name: create_brick
           run: tendrl.gluster.objects.GlusterBrick.atoms.Create
           type: Create
           uuid: 242f6190-9b37-11e6-950d-a24fc0d9649d
@@ -180,6 +190,9 @@ namespace.gluster:
           type: String
         node_id:
           help: node_id of the node on which brick recides
+          type: String
+        volume_name:
+          help: volume_name of the volume to which brick belongs
           type: String
         mount_path:
           help: mount point of the brick
@@ -210,7 +223,7 @@ namespace.gluster:
           type: String
       enabled: true
       value: clusters/$TendrlContext.integration_id/nodes/$TendrlContext.node_id/GlusterBricks/$GlusterBrick.brick_name
-      list:  clusters/$TendrlContext.integration_id/nodes/$TendrlContext.node_id/GlusterBricks/
+      list:  clusters/$TendrlContext.integration_id/nodes/$TendrlContext.node_id/GlusterBricks/all
       help: gluster bricks that can be used for gluster volumes
     Volume:
       atoms:
@@ -238,8 +251,6 @@ namespace.gluster:
             mandatory:
               - Volume.volname
               - Volume.vol_id
-            optional:
-              - Volume.format_bricks
           name: delete_volume
           run: gluster.objects.Volume.atoms.Delete
           type: Delete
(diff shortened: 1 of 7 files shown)


nnDarshan:
@brainfunked @r0h4n @shtripat @nthomas-redhat Please review.
To Be VERIFIED

tendrl/gluster_integration/objects/gluster_brick_path/__init__.py
```diff
@@ -0,0 +1,23 @@
+from tendrl.commons import objects
+
+
+class GlusterBrickDir(objects.BaseObject):
```
brainfunked: Should this not be DefaultBrickPath?

tendrl/gluster_integration/objects/volume/atoms/create/__init__.py
```diff
@@ -72,6 +72,38 @@ def run(self):
...
+                    ip = brick.keys()[0]
+                    brick_path = brick.values()[0]
+                    node_id = NS._int.client.read("indexes/ip/%s" % ip).value
+                    NS._int.wclient.delete(
+                        ("clusters/%s/nodes/%s/GlusterBricks/free/%s") % (
+                            NS.tendrl_context.integration_id,
+                            NS.node_context.node_id,
+                            brick_path.replace("/","_")
+                        ),
+                        recursive=True
```
brainfunked: Is `recursive=True` required? Would be safer to not include it if it isn't absolutely necessary.

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -60,6 +60,10 @@ def _run(self):
                     SyncObject(data=json.dumps(raw_data))
                 sync_object.save()
 
+                gluster_brick_dir = NS.gluster.objects.\
+                                     GlusterBrickDir()
+                gluster_brick_dir.save()
```
brainfunked: The value of the default brick directory should be set just once, during the cluster import. That way, it would be applicable to both the import and create scenarios. The value would not need to be synced ever again. Essentially, it becomes read-only. If needed, in the future, any additional brick directories (manually added by the user or based on detection from the initial cluster import) could be added to a separate list object.

tendrl/gluster_integration/objects/gluster_brick/__init__.py
```diff
@@ -20,6 +20,7 @@ def __init__(
         disk_type=None,
         disk_count=None,
         stripe_size=None,
+        volume_name=None
```
r0h4n: Dude, you are missing a comma here, this is gonna fail

nnDarshan: ack

tendrl/gluster_integration/sds_sync/__init__.py
nnDarshan: as mentioned above :
My idea was to use the same object to store additional attributes related to gluster brick dir. you were hinting at having a list of user defined brick paths in future, which would also come under this object as a new attribute like: "user_brick_paths"
So in that case user might change the brick paths in between so it has to be synced. hence its placed here.
Please provide your thoughts

tendrl/gluster_integration/objects/gluster_brick_path/__init__.py
nnDarshan: My idea was to use the same object to store additional attributes related to gluster brick dir. you were hinting at having a list of user defined brick paths in future, which would also come under this object as a new attribute like: "user_brick_paths"
Please provide your thoughts.

tendrl/gluster_integration/objects/volume/atoms/create/__init__.py
nnDarshan: ack

tendrl/gluster_integration/sds_sync/__init__.py
brainfunked: That is true. But I was thinking of having a clear separation between tendrl defaults and user configuration. Hence the defaults in a read-only object with the user/system configuration in a separate object that's subjected to sync. @r0h4n what do you think?

r0h4n: We should have separation between tendrl internal config vs user config, ill file a bug for this on specs after next build and figure out how much work is needed.

For now, this seems fine. @nnDarshan you can create and save GlusterBrickDir in the first sync only and skip it for any other syncs

tendrl/gluster_integration/objects/definition/gluster.yaml
```diff
@@ -25,6 +25,8 @@ namespace.gluster:
       uuid: 1951e821-7aa9-4a91-8183-e73bc8275b8e
```
nnDarshan: ack

tendrl/gluster_integration/sds_sync/__init__.py
nnDarshan: ack

nnDarshan:
fixes https://github.com/Tendrl/gluster-integration/issues/257

nnDarshan:
fixes https://github.com/Tendrl/gluster-integration/issues/258

tendrl/gluster_integration/objects/gluster_brick/atoms/create/__init__.py
```diff
@@ -87,8 +88,13 @@ def run(self):
...
                         **args
-                    ).save(update=False)
-
+                    ).save()
+                    free_brick_key = "clusters/%s/nodes/%s/GlusterBricks/free/%s" % (
+                        NS.tendrl_context.integration_id,
+                        val["node_id"],
+                        brick_name.replace("/","_")
+                    )
+                    NS._int.wclient.write(free_brick_key, "")
```
r0h4n: why not use delete instead of write?

tendrl/gluster_integration/objects/gluster_brick_path/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(
+        self,
+        default_brick_dir=None,
+        *args,
+        **kwargs
+    ):
+        super(GlusterBrickDir, self).__init__(*args, **kwargs)
+
+        self.default_brick_dir = default_brick_dir if default_brick_dir else \
```
r0h4n: Hey, why do you need object to encapsulate a config item?

This default_brick_dir can be accessed directly anywhere via NS.config.data.get('gluster_bricks_dir', "/tendrl_gluster_bricks")

tendrl/gluster_integration/objects/gluster_brick/atoms/create/__init__.py
nnDarshan: We have following things under gluster_bricks:
- free: containg just the name of free bricks
- used: containg just the name of used bricks
- all: contains all the details of all the bricks
free and used are kind of index.
Here when I create brick I update the detail under "all" and free section

tendrl/gluster_integration/objects/gluster_brick_path/__init__.py
nnDarshan: UI needs the default path as an attribute under cluster

r0h4n:
@nnDarshan verified?

nnDarshan:
@r0h4n, its verified

PR #263, opened 2017-05-19
Fix Volume create flow.

nnDarshan:
Volume create flow had following 2 issues:
1. Volume was explicitly started after create. We need not
do this as gdeploy starts a volume when created. This was
causing a failure as we tried starting a started volume.
2. While marking bricks as used, the path was not correctly
composed.

This patch fixes above 2 issues


nnDarshan:
@r0h4n @shtripat @nthomas-redhat Please review

r0h4n:
Verified?

nnDarshan:
@r0h4n yes verified

PR #275, opened 2017-05-25
Fix the type of message

nnDarshan:
Change the type of message from error to info.


nnDarshan:
@r0h4n @shtripat @nthomas-redhat Please review

PR #280, opened 2017-06-01
Added logic to lock the bricks while volume operations

shtripat:
Bug-URL: https://bugzilla.redhat.com/1457659

### DIFF
--- a/tendrl/gluster_integration/objects/volume/atoms/create/__init__.py
+++ b/tendrl/gluster_integration/objects/volume/atoms/create/__init__.py
@@ -1,10 +1,15 @@
 import etcd
+import json
 import subprocess
 
 from tendrl.commons.event import Event
 from tendrl.commons.message import Message
 from tendrl.commons import objects
+from tendrl.commons.objects import AtomExecutionFailedError
 from tendrl.gluster_integration.objects.volume import Volume
+from tendrl.gluster_integration.objects.volume import Volume
+from tendrl.gluster_integration.objects.volume.atoms import \
+    brick_lock_utils
 
 
 class Create(objects.BaseAtom):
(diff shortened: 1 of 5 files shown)


shtripat:
@nnDarshan plz review

tendrl/gluster_integration/objects/volume/atoms/create/__init__.py
```diff
@@ -41,6 +41,43 @@ def run(self):
...
+                except etcd.EtcdKeyNotFound:
+                    # Find node id using hostname
+                    nodes = NS._int.client.read("nodes/")
+                    for node in nodes.leaves:
+                        hostname = NS._int.client.read(
+                            "%s/NodeContext/fqdn" % node.key).value
+                        if hostname == ip:
+                            node_id = node.key.split("/")[-1]
+                key = "nodes/%s/NodeContext/fqdn" % node_id
+                host = NS._int.client.read(key).value
```
GowthamShanmugam: @shtripat i feel you can place this code below 52th line because if the ip is already a hostname  then this step is like repetitive.

shtripat: I need `node_id` for forming the key to be read from `/nodes/{id}/NodeContext/fqdn`. If `indexes/ip/{id}` available it takes `node_id` from there, otherwise from nodes list. Anyway in all the scenarios `node_id` would be available so keeping it here also is fine.

GowthamShanmugam: you can move line 61 and 62 into try
key = "nodes/%s/NodeContext/fqdn" % node_id
host = NS._int.client.read(key).value

because here you are identifying hostname, but in exception block already host name is there (ip is hostname). 
and you can assign host = ip in exception block is enaf

nnDarshan:
@shtripat Can you pls help me understand, where are you checking if bricks are locked before volume is created. I just see, code for setting lock before brick creation. Am I missing something

shtripat:
@nnDarshan thats correct. we would need to check pre-existence of any lock and if so, break the flow saying few of the bricks are already locked for volume creation. will update the PR accordingly. Anyway its not yet verified :)

tendrl/gluster_integration/objects/volume/atoms/create/__init__.py
shtripat: @GowthamShanmugam I think I got your point. Let me verify the changes once and would do the needful.

shtripat: @GowthamShanmugam so the volume creation request can have brick details in below two forms

```
"Volume.bricks": [
            [
                {"12.34.45.56":"/mnt/sample5/sampleBrick1"},
                {"12.34.45.56":"/mnt/sample4/sampleBrick1"}
            ]
],
```

or 

```
"Volume.bricks": [
            [
                {"fqdn1":"/mnt/sample5/sampleBrick1"},
                {"fqdn2":"/mnt/sample4/sampleBrick1"}
            ]
],
```

Suppose scenario-1 is followed and IP is provided. In that case we look at `/indexes/ip/{ip}` and get the node_id from there.

If scenario-2 was followed the key `/indexes/ip/{fqdn1}` will not be found and key not found exception occurs. In that case I loop through the list of nodes and wherever the node fqdn matches, that node's node_id I record to use.

I dont any issue in this logic.

shtripat:
@nnDarshan plz review

shtripat:
>> We should unlock the bricks if volume creation fails (so that these bricks can be used to create volume
>> again ) else these will be blocked for ever

Agree. Will do that.

>> It would be great if you can modularize locking and unlocking as this can be reused for expand volume as >> well

Will make check for brick locked as pre-run atom. Also would move locking and un-locking logic to a common place so that it can be re-used across flows.

shtripat:
Changes not yet verified. Will update once verified.

shtripat:
Verified the create volume flow. Expand volume flow to be verified.

shtripat:
Verified the expand volume flow as well and works as expected.

shtripat:
@nnDarshan do you need this for expand cluster flow as well? I feel yes.

r0h4n:
Parking this one for now

Issue #279, opened 2017-05-29
EtcdLockExpired exception intermittent for nodes in a Gluster cluster; impacts out-of-band delete operations.


GowthamShanmugam:
I have 4 nodes under gluster cluster, i have checked sds sync for all nodes in cluster, problem is node sync is not happening for all nodes, sometimes 2 nodes updating all sync details but other two nodes are giving etcd.EtcdLockExpired. Both two node are always giving this exception. if i restart gluster-integration then some times 3 nodes are updating fine and one node is always giving EtcdLockExpired exception. Because of this we can't give time out for out of band delete. 

The Volume details are updated from any nodes from same cluster, so this problem is hidden till now because of this. Problem is for brick details update. Brick detail are updated node wise So all brick details will never come in etcd.

GowthamShanmugam:
@r0h4n @nnDarshan @shtripat @nthomas-redhat @sankarshanmukhopadhyay

GowthamShanmugam:
@nnDarshan @r0h4n @sankarshanmukhopadhyay increasing seconds in 
   lock.acquire(blocking=True,lock_ttl=200)
                    if lock.is_acquired:
                        # renewing the lock
                        lock.acquire(lock_ttl=200)

is solving problem, for 4 machines 200 seconds are ok, what about more than 10 machines?

nnDarshan:
Myself @shtripat and @GowthamShanmugam had a discussion about a way to resolve this.

Since the ttl that we set depends on the number of nodes in the gluster cluster. Its not possible to arrive at a correct lock_ttl duration.
The ttl was introduced to avoid the issue where, a gluster_integration process is stopped while it has acquired the lock and not released it yet. This used to cause the lock held by this process for ever.

So instead of using ttl to solve the above problem, we can either: a) explicitly release the lock when we handle the SIGTERM signal in gluster_integration process OR b) Handle it during the start of gluster_integration service by removing the lock from etcd explicitly.
I feel solution a) is better.
@rohan please provide your thoughts

shtripat:
@nnDarshan I am in favor of approach-1. Is there a way we know what all locks are held by this agent and clean those only?

GowthamShanmugam:
@r0h4n need your suggestion

GowthamShanmugam:
#266 is a problem because of this issue

GowthamShanmugam:
@r0h4n @nnDarshan @shtripat 
[...]

Issue #266, opened 2017-05-22
Handle out of band deletes


nthomas-redhat:
Out of band deletes are not handled in gluster-integration
Tendrl has capabililty. any object (like volume) that wants to detect out of band deletes will need to add a TTL to its save()

GowthamShanmugam:
@r0h4n we can close this issue

PR #285, opened 2017-06-05
Out of band deletion for volume, brick and peer

GowthamShanmugam:
tendrl-bug-id: Tendrl/gluster-integration#266
tendrl-bug-id: Tendrl/gluster-integration#279

Lock code is removed because ttl will tack care of out of band deletion 
Signed-off-by: root <gshanmug@redhat.com>

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -68,17 +68,6 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
 
         while not self._complete.is_set():
             try:
-                # Acquire lock before updating the volume details in etcd
-                # We are blocking till we acquire the lock.
-                # the lock will live for 60 sec after which it will be released.
-                lock = etcd.Lock(NS._int.wclient, 'volume')
-                try:
-                    lock.acquire(blocking=True,lock_ttl=60)
-                    if lock.is_acquired:
-                        # renewing the lock
-                        lock.acquire(lock_ttl=60)
-                except etcd.EtcdLockExpired:
-                    continue
                 gevent.sleep(
                     int(NS.config.data.get("sync_interval", 10))
                 )
@@ -110,7 +99,8 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                                     hostname=peers['peer%s.primary_hostname' % index],
                                     state=peers['peer%s.state' % index]
                                 )
-                            peer.save()
+                            peer.save(ttl=int(
+                                NS.config.data.get("sync_interval", 10)+15))
                             index += 1
                         except KeyError:
                             break
@@ -199,7 +189,8 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                                     'volume%s.snapd_svc.inited' % index
                                 ],
                             )
-                            volume.save()
+                            volume.save(ttl=int(
+                                NS.config.data.get("sync_interval", 10)+15))
                             rebalance_details = NS.gluster.objects.RebalanceDetails(
                                 vol_id=volumes[
                                     'volume%s.id' % index
@@ -226,7 +217,8 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                                     'volume%s.rebalance.data' % index
                                 ],
                             )
-                            rebalance_details.save()
+                            rebalance_details.save(ttl=int(
+                                NS.config.data.get("sync_interval", 10)+15))
                             b_index = 1
                             # ipv4 address of current node
                             try:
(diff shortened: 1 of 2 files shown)


GowthamShanmugam:
@r0h4n @shtripat @nnDarshan please review

GowthamShanmugam:
Verified

GowthamShanmugam:
@shtripat yes i tested i have stopped gluster-integration for 200 sec, then all volume and peer and brick details are gone, then i start gluster-integration again then all entries present, with ttl

shtripat:
Forgot to mention earlier, you need to verify the delete volume flow as well and see there is no issue in the same. Also I feel explicit deletion of volume entry while flow should be removed as we discussed earlier and it would be deleted as part of next ttl.

GowthamShanmugam:
@shtripat i have verified volume deletion from command line and detach peer from gluster cluster from command line also

shtripat:
OK. Changes look fine to me.

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -35,17 +35,6 @@ def _run(self):
...
             try:
-                # Acquire lock before updating the volume details in etcd
-                # We are blocking till we acquire the lock.
-                # the lock will live for 60 sec after which it will be released.
-                lock = etcd.Lock(NS._int.wclient, 'volume')
-                try:
-                    lock.acquire(blocking=True,lock_ttl=60)
-                    if lock.is_acquired:
-                        # renewing the lock
-                        lock.acquire(lock_ttl=60)
```
r0h4n: @nnDarshan this is gonna affect DeleteVolume flow, can you review this PR

GowthamShanmugam:
I have tested this PR with 6 gluster cluster nodes, i feel it works fine, i have tested volume deletion also.

tendrl/gluster_integration/sds_sync/__init__.py
r0h4n: @nnDarshan can you comment on this? we need to get this moving

nnDarshan:
@GowthamShanmugam Remove the lock while deleting volume https://github.com/Tendrl/gluster-integration/blob/develop/tendrl/gluster_integration/objects/volume/atoms/delete/__init__.py#L88
Now since we are removing locking from sds_sync there is no point in locking only during sync.

nnDarshan:
@shtripat @r0h4n @GowthamShanmugam With this change the behavior of volume delete will be as follows, Gluster integration will fire an explicit delete to etcd after actual volume is deleted. But some times, due to the sync from gluster integration in other nodes the volume detail might again appear. But eventually after ttl this volume will go off from etcd.

I guess everyone is okay with this behavior.

GowthamShanmugam:
@nnDarshan i will remove lock from delete volume also

r0h4n:
Please resolve merge conflicts

GowthamShanmugam:
@r0h4n @nnDarshan all changes done, here we can use only ttl for out of band deletion, because brick is node wise, peer also current node peer id never come only other gluster nodes will come, here i feel ttl is good option

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -108,7 +97,7 @@ def _run(self):
                                     hostname=peers['peer%s.primary_hostname' % index],
                                     state=peers['peer%s.state' % index]
                                 )
-                            peer.save()
+                            peer.save(ttl=200)
```
r0h4n: the ttl value should be calculated  using local config flag `sync_interval` + 15 seconds 

check this for more details https://github.com/Tendrl/gluster-integration/pull/313

GowthamShanmugam: ok, i will change it

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -15,6 +15,9 @@
 from tendrl.gluster_integration import ini2json
 from tendrl.commons.utils.time_utils import now as tendrl_now
 
+# For 10 min (sync_time(10) + 590)
+SYNC_TTL = 590
```
GowthamShanmugam: sync_time +15 is not enough for sds sync because it depends on number of bricks and volumes and peers. If number of volumes are very high then some volumes are expired before sync happens. I have disused @shtripat about this so we decided  to set 10 min for sds sync

r0h4n: set this to SYNC_TTL=250

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -110,7 +102,8 @@ def _run(self):
                                     hostname=peers['peer%s.primary_hostname' % index],
                                     state=peers['peer%s.state' % index]
                                 )
-                            peer.save()
+                            peer.save(ttl=int(
+                                NS.config.data.get("sync_interval", 10)) + SYNC_TTL)
```
r0h4n: Calculate `int(NS.config.data.get("sync_interval", 10)) + SYNC_TTL)` only once in this loop

GowthamShanmugam: ahh correct :) i  will do that

tendrl/gluster_integration/sds_sync/__init__.py
GowthamShanmugam: ok

GowthamShanmugam:
@r0h4n @shtripat changes done

Issue #286, opened 2017-06-06
Raise alerts for volume and brick status change.


nnDarshan:
Gluster integration should raise alerts when the status of either brick or volume changes

nnDarshan:
This has been implemented

PR #282, opened 2017-06-02
Change the structure of gluster bricks

nnDarshan:
This patch consolidates the gluster brick structure by
haveing all the brick details under single place in etcd
and linking it from other places where its needed.

tendrl-bug-id: Tendrl/gluster-integration#278
Signed-off-by: nnDarshan <darshan.n.2024@gmail.com>

PR #287, opened 2017-06-06
Trigger alert for vol and brick failures

nnDarshan:
This patch sends alerts when the status of volume/bricks changes. Both from healthy to unhealthy and vice versa

tendrl-bug-id: Tendrl/gluster-integration#286


nnDarshan:
@shtripat @r0h4n @nthomas-redhat @anmolbabu Please review.
Have tested by checking an event with  priority "notice" appears in the "/messages/events" when status of volume or brick changes

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -20,6 +23,37 @@ def __init__(self):
         super(GlusterIntegrationSdsSyncStateThread, self).__init__()
         self._complete = gevent.event.Event()
 
+    def _emit_event(self, resource, curr_value, msg):
+        alert = {}
+        alert['source'] = 'tendrl-gluster-integration'
+        alert['pid'] = os.getpid()
+        alert['timestamp'] = tendrl_now().isoformat()
```
shtripat: This should be `time_stamp` as per latest testing with alerting module.

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -20,6 +23,37 @@ def __init__(self):
...
+        alert['alert_type'] = 'status'
+        severity = "info"
+        if curr_value == "Stopped":
+            severity = "critical"
+        alert['severity'] = severity
+        alert['resource'] = resource
```
shtripat: Resource name should be something like `volume_status` and the same should be populated as `handles` field of a handler which needs to added under `tendrl-alerting` module as `handlers/cluster/volume_status_handler.py`

You can refer https://github.com/Tendrl/alerting/pull/78 for more details.

If required a separate handler to be written for brick status changes.

tendrl/gluster_integration/objects/brick/__init__.py
```diff
@@ -4,34 +4,60 @@
 class Brick(objects.BaseObject):
```
r0h4n: We are not landing the Gluster Brick changes https://github.com/Tendrl/gluster-integration/pull/282 
until the UI folks ack on that, So please remove changes related to that and only keep the alert triggers

tendrl/gluster_integration/objects/brick/atoms/create/__init__.py
```diff
@@ -1,11 +1,10 @@
 from tendrl.commons.event import Event
 from tendrl.commons.message import Message
 from tendrl.commons import objects
-from tendrl.gluster_integration.objects.gluster_brick import GlusterBrick
+from tendrl.gluster_integration.objects.brick import Brick
```
r0h4n: Please dont mix up this PR and https://github.com/Tendrl/gluster-integration/pull/282 , the latter one requires UI work completion and the current PR needs to be merged without PR 282

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -20,6 +23,37 @@ def __init__(self):
         super(GlusterIntegrationSdsSyncStateThread, self).__init__()
         self._complete = gevent.event.Event()
 
+    def _emit_event(self, resource, curr_value, msg):
+        alert = {}
+        alert['source'] = 'tendrl-gluster-integration'
```
r0h4n: use NS.publisher_id

tendrl/gluster_integration/sds_sync/__init__.py
r0h4n: The resource should be the name of object as defined in Tendrl definitions i.e. Volume and the other part of resource should be the attribute on which the alert is triggered

Volume.status

anmolbabu: @r0h4n The alerting module expects it as volume_status or volume_utilization so basically its the combination of <entity_type>_<alert_type> where alert_type is either status or utilization

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -20,6 +23,37 @@ def __init__(self):
...
+        )
+        alert['node_id'] = NS.node_context.node_id
+        if not NS.node_context.node_id:
+            return
+        Event(
+            Message(
+                "notice",
+                "alerting",
+                {'message': json.dumps(alert)},
+                node_id=NS.node_context.node_id
```
anmolbabu: This is not required it will be automatically taken by Message class..

tendrl/gluster_integration/sds_sync/__init__.py
r0h4n: The alerting module does need to know about the object definition and attributes. Tendrl will be generating alerts on specific attributes of a Tendrl object.

But if you are expecting a underscore between the entity/object and alert/attribute we can continue with below scheme

<object>_<attribute>

works?

anmolbabu: Yes .. volume_status works with alerting as it stands today...

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -187,8 +248,67 @@ def _run(self):
...
+                                                    'volume%s.brick%s.path' % (
+                                                        index, b_index
+                                                    )
+                                                ],
+                                                volumes['volume%s.name' % index],
+                                                stored_brick_status,
+                                                current_status
+                                            )
+                                            self._emit_event(
+                                                volumes['volume%s.name' % index],
```
anmolbabu: for cases like brick where info about its volume and cluster both are important, please have an additional parameter under tags called the plugin_instance and have its value in accordance with https://github.com/Tendrl/node-monitoring/blob/develop/tendrl/node_monitoring/plugins/tendrl_glusterfs_brick_utilization.py#L286

Now, the reason for not having additional fields like vol_name, brick_path extra under the dict tags is , that collectd does not allow us to have custom additional fields in tags as the tags attribute for collectd generated alerts come directly from collectd based on how the plugin is configured(an example is the above link) and only a few reserved fields can be played around with which leaves plugin_instance as the best attribute choice left...
So if we decide to have custom fields in our(tendrl generated) alerts unless absolutely necessary will render the alerting module validations and alert clearing logic less generic...

nnDarshan:
Am taking care of all comments in this patch: https://github.com/Tendrl/gluster-integration/pull/288/
So closing this

Issue #296, opened 2017-06-14
The object creation job shouldnt be marked as completed till the object is populated in central store


shtripat:
The object creation jobs like volume/brick etc should not be marked completed till the object reflects in etcd

shtripat:
PR merged

PR #297, opened 2017-06-14
Added post checks to check if object are available in etcd post create

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#296

### DIFF
--- a/tendrl/gluster_integration/objects/volume/atoms/check_volume_available/__init__.py
+++ b/tendrl/gluster_integration/objects/volume/atoms/check_volume_available/__init__.py
@@ -1,0 +1,51 @@
+import etcd
+import gevent
+
+from tendrl.commons.event import Event
+from tendrl.commons.message import Message
+from tendrl.commons import objects
+from tendrl.commons.objects import AtomExecutionFailedError
+from tendrl.gluster_integration.objects.volume import Volume
+
+
+class CheckVolumeAvailable(objects.BaseAtom):
+    def __init__(self, *args, **kwargs):
+        super(CheckVolumeAvailable, self).__init__(*args, **kwargs)
+
+    def run(self):
+        retry_count = 0
+        while True:
+            try:
+                volumes = NS._int.client.read(
+                    "clusters/%s/Volumes" % NS.tendrl_context.integration_id
+                )
+            except etcd.EtcdKeyNotFound:
+                # ignore as no volumes available till now
+                pass
+
+            if volumes:
+                for entry in volumes.leaves:
+                    volume = Volume(vol_id=entry.key.split("Volumes/")[-1]).load()
+                    if volume.name == self.parameters['Volume.volname']:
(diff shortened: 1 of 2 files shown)


shtripat:
@r0h4n @nnDarshan plz review

tendrl/gluster_integration/objects/volume/atoms/check_volume_available/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+
+class CheckVolumeAvailable(objects.BaseAtom):
+    def __init__(self, *args, **kwargs):
+        super(CheckVolumeAvailable, self).__init__(*args, **kwargs)
+
+    def run(self):
+        retry_count = 0
+        while True:
+            try:
+                volumes = NS._int.client.read(
```
r0h4n: use object.load() , we need to ensure all attributes inside the object are also present as provided by user

Because a Volume is only really fully available when you can load() it


Also, just load the actual volume instead of iterating over every volume

tendrl/gluster_integration/objects/volume/atoms/check_volume_available/__init__.py
```diff
@@ -0,0 +1,52 @@
+import etcd
+import gevent
+import subprocess
```
r0h4n: subprocess not used

tendrl/gluster_integration/objects/volume/atoms/check_volume_available/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+                pass
+
+            if volumes:
+                for entry in volumes.leaves:
+                    volume = Volume(vol_id=entry.key.split("Volumes/")[-1]).load()
+                    if volume.name == self.parameters['Volume.volname']:
+                        return True
+
+            retry_count += 1
+            gevent.sleep(1)
```
r0h4n: Increase the timeout to 10 mins, we are going to allow configurable sds/node sync intervals in near future

tendrl/gluster_integration/objects/volume/atoms/check_volume_available/__init__.py
shtripat: Will remove it

tendrl/gluster_integration/objects/volume/atoms/check_volume_available/__init__.py
shtripat: As discussed earlier, while volume creation we have volume name only available. This post run atom is for create volume flow. To load a specific volume we need volume uuid which is not available here.

tendrl/gluster_integration/objects/volume/atoms/check_volume_available/__init__.py
shtripat: done

Issue #312, opened 2017-06-19
Configurable gluster sds sync interval


r0h4n:
The interval at which the gluster sds state is synced to tendrl central store should be user configurable with default provided by Tendrl

User should be able to set the `sync_interval` (seconds) in local config `/etc/tendrl/gluster-integration/gluster-integration.conf.yaml`

PR #313, opened 2017-06-19
Exposed SDS sync interval as a configuration

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#312


shtripat:

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -79,7 +79,9 @@ def _run(self):
                         lock.acquire(lock_ttl=60)
                 except etcd.EtcdLockExpired:
                     continue
-                gevent.sleep(10)
+                gevent.sleep(
+                    int(NS.config.data.get("sync_interval", 10))
```
r0h4n: This is not the only place that should use `sync_interval`

All the out of band object .save() with ttl should use ttl value by adding 10-15 seconds to `sync_interval`

shtripat: Should I grep for ttl and whenever there is ttl being set for the entities, change it to sync_interval + 15 ?

r0h4n: I am ok with whatever approach you take for this

shtripat: Actually at the moment, there are no ttl set for any entities in gluster-integration. Later if @GowthamShanmugam 's out of box delete PR needs ttl, he can use the logic to set the value as `sync_interval + 15`.

@GowthamShanmugam ^^^

PR #322, opened 2017-06-27
Added logic to sync cluster network details while import

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#321

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -10,6 +10,7 @@ from tendrl.gluster_integration.sds_sync import brick_utilization
 from tendrl.commons.event import Event
 from tendrl.commons.message import ExceptionMessage, Message
 from tendrl.commons.utils import cmd_utils
+from tendrl.commons.utils import etcd_utils
 
 from tendrl.commons import sds_sync
 from tendrl.gluster_integration import ini2json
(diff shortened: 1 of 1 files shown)


shtripat:
Not yet verified

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -66,6 +66,29 @@ def _run(self):
                             GlusterBrickDir()
         gluster_brick_dir.save()
 
+        try:
+            NS.tendrl.objects.ClusterNetworkDetails(
```
r0h4n: Use `NS.tendrl.objects.Cluster` instead of current

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -66,6 +66,29 @@ def _run(self):
...
+                    "nodes/%s/Networks" % NS.node_context.node_id
+                )
+                # TODO: (team) this logic needs to change later
+                # multiple networks supported for gluster use case
+                node_network = NS.tendrl.objects.NodeNetwork(
+                    interface=node_networks.leaves[0].value
+                ).load()
+                NS.tendrl.objects.ClusterNetworkDetails(
+                    integration_id=NS.tendrl_context.integration_id,
+                    public_network=public_network,
```
r0h4n: var `public_network` and `cluster_network` is not defined anywhere and yet it is used here

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: Done

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: Corrected

shtripat:
Verified with cluster sync scenario and adds the cluster_network properly.

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -66,6 +66,29 @@ def _run(self):
                             GlusterBrickDir()
         gluster_brick_dir.save()
 
+        try:
+            NS._int.client.read(
```
r0h4n: Use https://github.com/Tendrl/commons/blob/develop/tendrl/commons/utils/etcd_utils.py#L15

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -66,6 +66,29 @@ def _run(self):
...
+                )
+                # TODO: (team) this logic needs to change later
+                # multiple networks supported for gluster use case
+                node_network = NS.tendrl.objects.NodeNetwork(
+                    interface=node_networks.leaves.next().key.split('/')[-1]
+                ).load()
+                cluster = NS.tendrl.objects.Cluster(
+                    integration_id=NS.tendrl_context.integration_id
+                ).load()
+                cluster.cluster_network = node_network.subnet
```
shtripat: In case of gluster as recommendation single network would be available on nodes and yes in that case node_network.subnet should be equal to cluster.cluster_network.

UI uses the cluster.cluster_network value to filter the nodes which could be added to this cluster. If node's subnet matches the value then they can be added to the cluster.

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: Done

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -66,6 +67,29 @@ def _run(self):
...
+                cluster.save()
+            except etcd.EtcdKeyNotFound:
+                # pass as we should try to sync network details next time
+                pass
```
r0h4n: log appropriate msg and raise this exception instead of pass

shtripat: Done

Issue #353, opened 2017-08-10
Enable volume profiling during sds sync


r0h4n:
- If NS.tendrl.objects.Cluster.enable_volume_profiling is "true" 
Then
During sds sync, turn on gluster volume profiling for all volumes.

- If NS.tendrl.objects.Cluster.enable_volume_profiling is "false" 
Then
During sds sync, turn off gluster volume profiling for all volumes

shtripat:
I understand this needs to be done while import cluster flow and ideally flag should be coming as parameters for the flow. Also doing this from all the gluster nodes is not required, so we should check if profiling is already enabled/disabled and take action accordingly based on value of `NS.tendrl.objects.Cluster.enable_volume_profiling`.

Comments??

shtripat:
This can be closed as changes are merged already

PR #352, opened 2017-08-10
Spec/enhance gluster cluster sync logic

shtripat:
Changes for gluster sds sync for additional details

tendrl-spec: Tendrl/specifications#170

PR #354, opened 2017-08-10
Added logic to enable / disable volume profiling

shtripat:
If cluster level flag is set to enable profiling, while syncing
the cluster volumes, profiling would be enabled for all the
volumes, else disabled.

tendrl-bug-id: Tendrl/gluster-integration#353

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -650,7 +650,9 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                     _cluster.save()
 
                 # check and enable volume profiling
-                self._enable_disable_volume_profiling()
+                if "provisioner/%s" % NS.tendrl_context.integration_id in \
+                    NS.node_context.tags:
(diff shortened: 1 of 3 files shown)


shtripat:
Not yet verified

shtripat:
Verified. But it needs load_all() fix from https://github.com/Tendrl/commons/pull/685.

r0h4n:
@Tendrl/tendrl-core @nthomas-redhat please approve

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -649,6 +649,9 @@ def _run(self):
                     _cluster.is_managed = "yes"
                     _cluster.save()
 
+                # check and enable volume profiling
+                self._enable_disable_volume_profiling()
```
r0h4n: This function should be called only by the gluster-integration with tag "provisioner/$integration_id"

shtripat: Done. will add a check for the same.

shtripat:
>> #352 would be required for this right?
No. This is independent and depends on load_all() PR only.

PR #368, opened 2017-08-22
Find rebalance status for all distribute volumes

nnDarshan:
Earlier rebalance status was calculated for only pure distribute volume. Now it has been edited to gather status for all types of distribute volume


tendrl/gluster_integration/sds_sync/rebalance_status.py
```diff
@@ -50,7 +50,7 @@ def sync_volume_rebalance_estimated_time(volumes):
 
 def sync_volume_rebalance_status(volumes):
     for volume in volumes:
-        if volume.vol_type == "Distribute":
+        if volume.vol_type.startswith("Distribute"):
```
shtripat: Can we say `if "Distribute" in volume.vol_type` ??

Issue #383, opened 2017-08-31
The enable volume profiling would fail if no volumes


shtripat:
In case there are no volumes in the cluster, the volumes list reported by `load_all()` would be `None` and so looping through it, would fail.

shtripat:
Closing as changes merged...

PR #384, opened 2017-08-31
Initailize the volumes list as empty list if no volumes

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#383


shtripat:
@r0h4n plz review

Issue #388, opened 2017-09-04
Change the brick detail path to be placed under host directory in etcd


nnDarshan:
Currently the brick details are saved at `clusters/<cluster-id>/bricks/all/<host:brick-dir-path>` this has to be changed to `clusters/<cluster-id>/bricks/all/<host>/<brick-dir-path>`. 

The reason for this change is API can easily get a list of all bricks under a host with the new structure.

nnDarshan:
relevant patch is merged

PR #390, opened 2017-09-05
Change the path where brick details are saved.

nnDarshan:
Currently the brick details are saved at:
clusters/<cluster-id>/bricks/all/<host:brick-dir-path>
This PR changes the path to:
clusters/<cluster-id>/bricks/all/<host>/<brick-dir-path>.

The reason for this change is API can easily get a list
of all bricks under a host with the new structure.

tendrl-bug-id: Tendrl/gluster-integration#388


nnDarshan:
@r0h4n @shtripat @brainfunked @anivargi @nthomas-redhat Please review

nnDarshan:
> Looks fine. Are we doing some specific read as well in other places? If so they also should be changed (if any)
@shtripat Thanks for pointing, will update accordingly

PR #392, opened 2017-09-06
Regression fix with emit_event

shtripat:
Fixed a regression issue.


shtripat:
@r0h4n this is a regression issue. Plz merge.

Issue #394, opened 2017-09-06
Save volume name as an attribute of brick object


nnDarshan:
Volume name has to be saved as an attribute under brick object, as it is needed to be shown in UI during brick listing.

nnDarshan:
relevant patch is merged

PR #395, opened 2017-09-06
Save volume same as an attribute of brick object

nnDarshan:
tendrl-bug-id: Tendrl/gluster-integration#394


nnDarshan:
@shtripat @r0h4n @nthomas-redhat @gnehapk Please review

PR #396, opened 2017-09-07
Added alerts for status changes

shtripat:
Alerts added for
- rebelance status changes
- geo-rep session status changes
- volume status changes
- brick status changes
- peer status changes

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -155,8 +155,41 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                                     hostname=peers[
                                         'peer%s.primary_hostname' % index
                                     ],
-                                    state=peers['peer%s.state' % index]
+                                    state=peers['peer%s.state' % index],
+                                    connected=peers['peer%s.connected' % index]
                                 )
+                            try:
+                                stored_peer_status = NS._int.client.read(
+                                    "clusters/%s/Peers/%s/connected" % (
+                                        NS.tendrl_context.integration_id,
+                                        peers['peer%s.uuid' % index]
+                                    )
+                                ).value
+                                current_status = peers[
+                                    'peer%s.connected' % index
+                                ]
+                                if stored_peer_status != "" and \
+                                    current_status != stored_peer_status:
+                                    msg = ("Status of peer: %s "
+                                           "changed from %s to %s") % (
+                                               peers[
+                                                   'peer%s.primary_hostname' %
+                                                   index
+                                               ],
+                                               stored_peer_status,
+                                               current_status)
+                                    instance = "peer_%s" % peers[
+                                        'peer%s.primary_hostname' % index
+                                    ]
+                                    event_utils.emit_event(
+                                        "peer_status",
+                                        current_status,
+                                        msg,
+                                        instance
+                                    )
+                            except etcd.EtcdKeyNotFound:
+                                pass
+
                             if not peer.exists():
                                 job_id = monitoring_utils.update_dashboard(
                                     peer.hostname,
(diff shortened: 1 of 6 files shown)


shtripat:

shtripat:
@nnDarshan @anmolbabu @GowthamShanmugam plz review

tendrl/gluster_integration/sds_sync/georep_details.py
```diff
@@ -36,6 +38,56 @@ def save_georep_details(volumes, index):
...
+                        volumes['volume%s.id' % index],
+                        session_id,
+                        pair_name
+                    )
+                ).value
+                pair_status = volumes[
+                    'volume%s.pair%s.status' % (index, pair_index)
+                ]
+                if fetched_pair_status != pair_status and \
+                    pair_status.lower() == 'faulty':
```
nnDarshan: the if block below looks very similar, can we club it to single block by having `pair_status.lower() in ['faulty','active', 'passive']:` and having place holder for new status

tendrl/gluster_integration/sds_sync/event_utils.py
```diff
@@ -1,39 +1,36 @@
...
+def emit_event(resource, curr_value, msg, instance):
     alert = {}
     alert['source'] = NS.publisher_id
+    alert['classification'] = 'cluster'
     alert['pid'] = os.getpid()
     alert['time_stamp'] = tendrl_now().isoformat()
-    alert['alert_type'] = 'status'
+    alert['alert_type'] = 'STATUS'
     severity = "INFO"
     if curr_value.lower() == "stopped":
```
nnDarshan: This was valid when we only had volume and brick, Now geo-rep/rebalance/peer is something else right, Can we modify this function to take "severity" as an argument and let caller of this utility fun decide about the severity ?

tendrl/gluster_integration/sds_sync/georep_details.py
shtripat: so the logic is like in case current status is `faulty`, the previous could be anything. But if the current status is `active`/`passive`, the old value we look for `faulty` only so I kept as separate conditions.

tendrl/gluster_integration/sds_sync/event_utils.py
shtripat: That's a good catch. Would do the required changes.

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -467,27 +499,30 @@ def sync_volumes(volumes, index, vol_options):
             try:
                 sbs = NS._int.client.read(
                     "clusters/%s/Bricks/all/"
-                    "%s/status" % (
+                    "%s/%s/status" % (
                         NS.tendrl_context.
                         integration_id,
-                        brick_name
+                        NS.node_context.fqdn,
+                        brick_name.replace(
```
r0h4n: @nnDarshan Is the brick name correctly derived here? As per our discussion on not having "/" for any attribute value

nnDarshan: @shtripat @r0h4n This is technically fine, but `brick_name.split(":_")[-1]` would suffice now.

tendrl/gluster_integration/sds_sync/georep_details.py
nnDarshan: Ack

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: Rabsed with latest master below and should be fine

tendrl/gluster_integration/sds_sync/event_utils.py
```diff
@@ -27,7 +21,7 @@ def emit_event(resource, curr_value, msg, instance):
         integration_id=NS.tendrl_context.integration_id,
         cluster_name=NS.tendrl_context.cluster_name,
         sds_name=NS.tendrl_context.sds_name,
-        fqdn=socket.getfqdn()
+        fqdn=NS.node_context.fqdn
     )
     alert['node_id'] = NS.node_context.node_id
     if not NS.node_context.node_id:
```
GowthamShanmugam: For cluster related alert node_id is not mandatory

r0h4n: We should make the node_id mandatory, Every alert even inside a cluster will come from some node_id

Issue #400, opened 2017-09-11
Few attributes in brick object is coming as unicode instead of string


nnDarshan:
Few attributes in brick object is coming as unicode instead of string, this is causing invalid json in API layer.

nnDarshan:
fix merged

PR #401, opened 2017-09-11
Change unicode attrs in brick obj to string.

nnDarshan:
If attributes are of unicode type, its causing issues
in API layer.

tendrl-bug-id: Tendrl/gluster-integration#400


nnDarshan:
@anivargi @r0h4n @nthomas-redhat @shtripat PLs review

Issue #413, opened 2017-09-13
Send volume name as well in update dashborad flow while add/delete brick


shtripat:
The monitoring-integration module needs volume name as well while add/delete brick event for updating dashboard. The same should be passed while invoking the flow from gluster-integration.

shtripat:
Closing this as required PR is merged

PR #414, opened 2017-09-13
Added volume name as well in resource while brick add/delete

shtripat:
While invoking update dashboard flow for brick add/delete, the
monitoring-integration needs volume name as well. Added the volume
name as part resource_name parameter of the flow in the form
`{vol_name}|{brick-host-name}:{brick-path}`

tendrl-bug-id: Tendrl/gluster-integration#413


shtripat:
@nnDarshan plz review

nnDarshan:
Looks good

Issue #418, opened 2017-09-14
Volume profiling flag is not set as instructed always


nthomas-redhat:
When the volume profiling is enabled already, this fails and flag is not appropritately in etcd

nnDarshan:
@shtripat Looks like the fix is merged. Could you please confirm and update accordingly

shtripat:
@nthomas-redhat plz do the needfil ^^^

PR #427, opened 2017-09-20
Added logic to check already started/stopped profiling for volumes

shtripat:
If profiling is already started for few volumes from CLI and we
import the cluster, by no means these volumes were marked as
profiling on in tendrl. Modified to tackle this scenario.

tendrl-bug-id: Tendrl/gluster-integration#418


shtripat:
@nthomas-redhat @r0h4n @nnDarshan plz review

PR #428, opened 2017-09-21
Updated to mark the volume state as down if stopped

shtripat:
If the volume status is stopped we mark the state as down rightaway.
Also not all the bricks are reported yet in central store of tendrl
from all the participating nodes of the cluster (i.e. the total no of
bricks reported till now is less than the brick_count field of volume),
the volume state remains as `unknown`.

So now the calculation logic of volume state triggers only if volume
is started and if not all the bricks updated yet in central store, the
volume state is marked as `unknown`.

tendrl-bug-id: Tendrl/monitoring-integration#23


shtripat:
@nthomas-redhat @r0h4n @nnDarshan plz review.

@atinmu please check if this solution is correct. To summarize logic is as below

1. If the volume is stopped, state of the volume is marked as `down`
2. If volume is started but not all the bricks are reported to etcd yet, the volume state remains as `unknown` and the moment all the bricks are reported based on type of the volume and its subvolume brick's status, the volume state value is calculated to `up`, `up(partial)` or `up(degraded)`

@atinmu your comments are required for guidelines.

Issue #433, opened 2017-09-25
Fix gluster native event handling.


nnDarshan:
gluster native events that are being ignored by gluster integration are not handled correctly, this is causing an exception in event listener. Which in turn is leading to an `500 internal server error` in glusterfs events.log

nnDarshan:
Fix patch is merged. Closing this

PR #434, opened 2017-09-25
Handle ignored native events properly.

nnDarshan:
gluster native events that are being ignored by
gluster integration are not handled correctly, this
is causing an exception in event listener. Which in
turn is leading to an `500 internal server error` in
glusterfs events.log

tendrl-bug-id: Tendrl/gluster-integration#433

### DIFF
--- a/tendrl/gluster_integration/message/callback.py
+++ b/tendrl/gluster_integration/message/callback.py
@@ -383,7 +383,7 @@ class Callback(object):
         native_event = NS.gluster.objects.NativeEvents(
             context,
             message=message,
-            severity="warning",
+            severity="info",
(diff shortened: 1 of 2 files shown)


nnDarshan:
@r0h4n @shtripat Please review

shtripat:
is it like it would show `Event Ignored` on UI for ignored ones?

nnDarshan:
> is it like it would show Event Ignored on UI for ignored ones?

@shtripat No, this will be returned to the gluster events api, when it posts to the event-listener webhook.

shtripat:
@nnDarshan ok. Looks fine otherwise.

Issue #437, opened 2017-09-28
Invoke update graphite flow of monitoring-integration while deletion of peer/vol/brick


shtripat:
This flow should be executed to make sure, the graphite data for these entities get archived on deletion event.

shtripat:
Closing as the required PRs are merged

PR #438, opened 2017-09-28
Invoke update graphite while deletion of peer/vol/brick

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#437

### DIFF
--- a/tendrl/gluster_integration/message/callback.py
+++ b/tendrl/gluster_integration/message/callback.py
@@ -421,6 +421,21 @@ class Callback(object):
             }
         )
 
+        job_id = monitoring_utils.update_graphite(
+            event['message']['host'],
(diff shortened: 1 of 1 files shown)


shtripat:
@r0h4n @nnDarshan @rishubhjain @GowthamShanmugam plz review

tendrl/gluster_integration/message/callback.py
```diff
@@ -421,6 +421,21 @@ def peer_disconnect(self, event):
             }
         )
 
+        job_id = monitoring_utils.update_graphite(
```
r0h4n: change as per https://github.com/Tendrl/commons/pull/741#discussion_r142022501

shtripat: Done

Issue #439, opened 2017-09-28
Set deleted flag while deletion of entities peer/volume/brick


shtripat:
Set this flag while events for deletion of peer/volume/brick. This is required to make sure once deletion event raised for an entity, the deleted flag should be set and UI/api should not list these entities.

rishubhjain:
@shtripat Should we do this for cluster as well?

shtripat:
Cluster status is a calculated value already and valid values are `healthy` and `unhealthy`. This is already taken care I remember.

shtripat:
Closing this as required PR are merged.

PR #440, opened 2017-09-28
Modified to set deleted flag while peer/vol/brick deletion

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#439

### DIFF
--- a/tendrl/gluster_integration/message/callback.py
+++ b/tendrl/gluster_integration/message/callback.py
@@ -9,6 +9,7 @@
 
 from tendrl.commons.utils import log_utils as logger
 from tendrl.commons.utils import monitoring_utils
+from tendrl.commons.utils import time_utils
 
 
 import time
@@ -391,7 +392,7 @@ class Callback(object):
 
     def peer_connect(self, event):
         job_id = monitoring_utils.update_dashboard(
-            event['message']['host'],
+            event['message']['peer'],
             RESOURCE_TYPE_PEER,
             NS.tendrl_context.integration_id,
             "add"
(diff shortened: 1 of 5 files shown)


shtripat:
@nthomas-redhat @r0h4n @nnDarshan @rishubhjain @GowthamShanmugam plz review

shtripat:
I have verified the volume delete scenario. Will verify the brick and peer delete as well and update.

tendrl/gluster_integration/message/callback.py
```diff
@@ -406,8 +407,15 @@ def peer_connect(self, event):
         )
 
     def peer_disconnect(self, event):
+        fetched_peers = NS.gluster.objects.Peer().load_all()
```
shtripat: I had a chat with Atin today morning and looks like we should not be removing the graphite data and mark peer deleted from monitoring integration in case of peer disconnect. This is a temporary event and peer might connect back soon.

PERR_DETACH is the event may be, to do these thing but at the moment this event is not handled in tendrl. I would suggest to remove the logic of removing graphite data and marking the peer as deleted on this event.

@atinmu can you please comment and confirm the same.
@Tendrl/tendrl-core need comments on this to remove the marking of peer as deleted and call of `delete_resource_from_graphite`.

atinmu: peer delete and peer disconnect are two different events. Disconnection can happen because of (a) if the node is down or (b) glusterd instance running on the node is down or (c) there's a n/w disconnect . In any of these cases, glusterd doesn't wipe off the peerinfo object from its peer list. On further connect event the peer status is set back to connect. So marking the peer deleted in case of a disconnect event doesn't sound correct to me.

shtripat:
Verified the volume delete and brick delete scenarios and sets the flag properly

Issue #446, opened 2017-10-16
Use statfs() on mounted volume or gfapi for getting the volume utilization


shtripat:
Currently uses a logic (lifted from gstatus) for calculation of volume utilization which is not a foolproof way of doing so. Also due to future changes or any internal changes within glusterfs would impact this logic in tendrl.

Rather, as suggested by glusterfs team, we should either be suing statfs() on mounted volume or gfapi call to get the volume utilization data.

shtripat:
@nnDarshan @nthomas-redhat @Tendrl/tendrl-core

shtripat:
Snippets from mail chain discussion with gluster team below


>> Yesterday while tendrl meeting, there was a concern raised by Martin Kudlej regarding mismatch >> between the volume utilization reported by tendrl and df command on underlying command.
>> The brick utilization data is returned by get-state command and we maintain in our central store.
>> The volume utilization data is aggregated based on a logic taken from gstatus in tendrl and 
>> updated to central store.
>> The volume utilization calculation logic in tendrl is present at [1]. Based on type of the volume,
>> the aggregation is done and updated.
>> Request you both to please check the logic and point out if any issues in our understanding and
>> if tendrl needs some changes for volume utilization calculation.

Reply from Atin
```Given Gluster is a distributed file system, it already offers the capability of the the brick details aggregation and offer the volume utilization capability by statfs() on the mount point of the volume. Individual brick's utilization can still be monitored through the get-state output or df on the brick. Having  such aggregation logic to be maintained for individual volume types is a pain and we should avoid that as much as possible. So tendrl just needs to send a statfs call (either through a temp mount or a gfapi call) for the volume.```

But @nthomas-redhat is of strong feeling that we should not mounting the volume and then getting the utilization details. Comments from @nthomas-redhat over gitter chat

```I don't think mounting volumes on provisioner node is the right way. All sorts of complications arise when nodes go down. Also asking admins to unmount before deleting is also is not user friendly. What I think is we need to push this to Gluster. Let Gluster provide us with right utilisation values at various levels.```
=========

shtripat:
@mkudlej @Tendrl/tendrl-qe @atinmu @amarts ^^^

mkudlej:
[...]

PR #450, opened 2017-10-23
Changed the logic to gfapi for volume utilization

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#446

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -107,8 +107,8 @@ setup(
         ],
     entry_points={
         'console_scripts': [
-            'tendrl-gluster-integration = '
-            'tendrl.gluster_integration.manager:main'
+            'tendrl-gluster-integration = tendrl.gluster_integration.manager:main',
+            'vol_utilization = tendrl.gluster_integration.sds_sync.vol_utilization:main'
         ]
     },
     include_package_data=True,
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,7 @@ setup(
     entry_points={
         'console_scripts': [
             'tendrl-gluster-integration = tendrl.gluster_integration.manager:main',
-            'vol_utilization = tendrl.gluster_integration.sds_sync.vol_utilization:main'
+            'tendrl-gluster-vol_utilization = tendrl.gluster_integration.sds_sync.vol_utilization:main'
         ]
     },
     include_package_data=True,
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,7 @@ setup(
     entry_points={
         'console_scripts': [
             'tendrl-gluster-integration = tendrl.gluster_integration.manager:main',
-            'tendrl-gluster-vol_utilization = tendrl.gluster_integration.sds_sync.vol_utilization:main'
+            'tendrl-gluster-vol-utilization = tendrl.gluster_integration.sds_sync.vol_utilization:main'
         ]
     },
     include_package_data=True,
--- a/tendrl-gluster-integration.spec
+++ b/tendrl-gluster-integration.spec
@@ -39,7 +39,6 @@ install -m  0755  --directory $RPM_BUILD_ROOT%{_sysconfdir}/tendrl/gluster-integ
 install -Dm 0644 tendrl-gluster-integration.service $RPM_BUILD_ROOT%{_unitdir}/tendrl-gluster-integration.service
 install -Dm 0644 etc/tendrl/gluster-integration/gluster-integration.conf.yaml.sample $RPM_BUILD_ROOT%{_datadir}/tendrl/gluster-integration/gluster-integration.conf.yaml
 install -Dm 644 etc/tendrl/gluster-integration/*.sample $RPM_BUILD_ROOT%{_datadir}/tendrl/gluster-integration/
-install -Dm 0755 tendrl/gluster-integration/sds_sync/vol_utilization.py %{_bindir}/vol_utilization
 
 %post
 systemctl enable tendrl-gluster-integration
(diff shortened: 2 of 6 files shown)


shtripat:
@nthomas-redhat @r0h4n @brainfunked @nnDarshan plz review

tendrl/gluster_integration/gfapi.py
```diff
@@ -0,0 +1,142 @@
...
+        ('f_fsid', ctypes.c_ulong),
+        ('f_flag', ctypes.c_ulong),
+        ('f_namemax', ctypes.c_ulong),
+        ('__f_spare', ctypes.c_int * 6),
+    ]
+
+
+def glfsInit(volumeId, host, port, protocol):
+    fs = _glfs_new(volumeId)
+
```
nnDarshan: Could you please add this check: https://github.com/oVirt/vdsm/blob/master/lib/vdsm/gluster/gfapi.py#L65

tendrl/gluster_integration/sds_sync/vol_utilization.py
```diff
@@ -0,0 +1,59 @@
...
+    return {'sizeTotal': float(total),
+            'sizeFree': float(free),
+            'sizeUsed': float(used)}
+
+
+def showVolumeUtilization(vname):
+    try:
+        data = gfapi.getVolumeStatvfs(vname)
+    except gfapi.GlusterLibgfapiException:
+        sys.stdout.write("CRITICAL: Failed to get the "
```
nnDarshan: if we write to stdout, will it report as err ? should we write to stderr ?
Also, it would be better to write the exception info as well because we will miss out why the script failed.

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -1,105 +1,28 @@
...
+        cmd = subprocess.Popen(
+            "vol_utilization %s" % volume.name,
+            shell=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            stdin=open(os.devnull, "r"),
+            close_fds=True
+        )
+        out, err = cmd.communicate()
+        if err == '':
```
nnDarshan: should we log on error ? so that, if volume utilization computation fails we know that there is some issue

tendrl/gluster_integration/gfapi.py
shtripat: Done

tendrl/gluster_integration/sds_sync/utilization.py
shtripat: Ah yes that would be good. Would add code for the same.

tendrl/gluster_integration/sds_sync/vol_utilization.py
shtripat: I had changed it. I think I forgot to push. Will do that.

tendrl-gluster-integration.spec
```diff
@@ -39,6 +39,7 @@ install -m  0755  --directory $RPM_BUILD_ROOT%{_sysconfdir}/tendrl/gluster-integ
 install -Dm 0644 tendrl-gluster-integration.service $RPM_BUILD_ROOT%{_unitdir}/tendrl-gluster-integration.service
 install -Dm 0644 etc/tendrl/gluster-integration/gluster-integration.conf.yaml.sample $RPM_BUILD_ROOT%{_datadir}/tendrl/gluster-integration/gluster-integration.conf.yaml
 install -Dm 644 etc/tendrl/gluster-integration/*.sample $RPM_BUILD_ROOT%{_datadir}/tendrl/gluster-integration/
+install -Dm 0755 tendrl/gluster-integration/sds_sync/vol_utilization.py %{_bindir}/vol_utilization
```
r0h4n: This shouldnt be handled here,

youll need to add an entry for "tendrl_gluster_vol_utilizatin: tendrl.gluster-integration.sds_sync/vol_utilization"

https://github.com/Tendrl/gluster-integration/blob/master/setup.py#L110

shtripat: Done. Will update

setup.py
```diff
@@ -107,8 +107,8 @@ def run(self):
         ],
     entry_points={
         'console_scripts': [
-            'tendrl-gluster-integration = '
-            'tendrl.gluster_integration.manager:main'
+            'tendrl-gluster-integration = tendrl.gluster_integration.manager:main',
+            'vol_utilization = tendrl.gluster_integration.sds_sync.vol_utilization:main'
```
r0h4n: Please change the cmd to "tendrl_gluster_vol_utilization" , "vol_utilization" is very ambiguous

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -44,7 +44,8 @@ def run(self):
         # To detect out of band deletes
         # refresh gluster object inventory at config['sync_interval']
         # Default is 260 seconds
-        SYNC_TTL = int(NS.config.data.get("sync_interval", 10)) + 250
+        SYNC_TTL = int(NS.config.data.get("sync_interval", 10)) +
+                   int(NS.config.data.get("ttl_delta", 600))
```
r0h4n: I would prefer "ttl_delta" to not be exposed to user. 

ttl_delta should be calculated on the count of volume and bricks

shtripat: Done. Will send another PR for the same.

PR #454, opened 2017-10-26
Add flow to update brick status as stopped.

nnDarshan:
This patch adds a new flow UpdateBrickStatus
that updates the status of bricks of a given
node as "Stopped". This flow is invoked for
nodes that are down.

tendrl-bug-id: Tendrl/monitoring-integration#169

### DIFF
--- a/tendrl/gluster_integration/flows/update_brick_status/__init__.py
+++ b/tendrl/gluster_integration/flows/update_brick_status/__init__.py
@@ -1,0 +1,23 @@
+from tendrl.commons import flows
+from tendrl.commons.utils import log_utils as logger
+
+
+class UpdateBrickStatus(flows.BaseFlow):
+    def run(self):
+        logger.log(
+            "info",
+            NS.publisher_id,
+            {"message": "starting flow for updating brick status"}
+        )
+        super(UpdateBrickStatus, self).run()
+        node = self.parameters.get("Node.fqdn")
+        cluster_id = self.parameters.get("TendrlContext.integration_id")
+        status = self.parameters.get("Brick.status")
+        self.update_brick_status(node, cluster_id, status)
+
+    def update_brick_status(self, node, cluster_id, status):
+        bricks = NS.gluster.objects.Brick(fqdn=node).load_all()
+        for brick in bricks:
+            brick.status = status
+            brick.save()
+        return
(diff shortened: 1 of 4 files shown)


tendrl/gluster_integration/flows/update_brick_status/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+            {"message": "starting flow for updating brick status"}
+        )
+        super(UpdateBrickStatus, self).run()
+        node = self.parameters.get("Node.fqdn")
+        cluster_id = self.parameters.get("TendrlContext.integration_id")
+        self.update_brick_status(node, cluster_id)
+
+    def update_brick_status(self, node, cluster_id):
+        try:
+            bricks = NS._int.client.read(
```
r0h4n: you should use the Brick object instead of raw read/writes to etcd

nnDarshan: Done

tendrl/gluster_integration/flows/update_brick_status/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+        )
+        super(UpdateBrickStatus, self).run()
+        node = self.parameters.get("Node.fqdn")
+        cluster_id = self.parameters.get("TendrlContext.integration_id")
+        self.update_brick_status(node, cluster_id)
+
+    def update_brick_status(self, node, cluster_id):
+        bricks = NS.gluster.objects.Brick(fqdn=node).load_all()
+        for brick in bricks:
+            brick.status = "Stopped"
```
shtripat: If status is `Stopped` hard coded why call the flow as `UpdateBrickStatus`?
Or else pass status as `Stopped` as parameter to the flow from the caller.

nnDarshan: ack

r0h4n: Yes, this flow needs to either "stop_brick" 

or

 "update_brick_status" and handle all statuses

nnDarshan:
@rohan Suggestions are taken care in latest Patch. Pls review. Also this patch is verified

PR #455, opened 2017-10-27
Calculate and set TTL for entities based on no of volumes

shtripat:


shtripat:
@Tendrl/tendrl-core @r0h4n @nnDarshan @nthomas-redhat plz review

Issue #459, opened 2017-11-02
georep session status changes


shtripat:
Make the required changes to accomodate all possible georep session status values as below

CREATED: Geo-replication session is established
STOPPED: Geo-replication session is stopped
ONLINE/UP: Geo-replication session (All bricks) are UP and Running
OFFLINE/DOWN: Geo-replication session (All bricks) are Down
PARTIAL: Geo-replication session - some bricks are Online and Some bricks are offline
PAUSED: all the pairs are in paused state

julienlim:
Ack the new 6 statuses for geo-repl sessions, and these should be reflected on the applicable Grafana dashboards. 

@shtripat

julienlim:
Related to https://github.com/Tendrl/gluster-integration/pull/460

shtripat:
Closing as required patches are merged

PR #460, opened 2017-11-02
Changes to accomodate all possible georep session status values

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#459


shtripat:
@cloudbehl @gnehapk this would need respective changes in UI as well.
@nnDarshan @nthomas-redhat @r0h4n plz review.
@julienlim Ack the required changes in UI for the same.

tendrl/gluster_integration/sds_sync/georep_details.py
```diff
@@ -188,19 +188,38 @@ def aggregate_session_status():
...
-                    if pair.status == "Faulty":
+                    if pair.status.lower() == "faulty":
                         faulty_count += 1
+                    elif pair.status.lower() == "created":
+                        created_count += 1
+                    elif pair.status.lower() == "stopped":
+                        stopped_count += 1
+                    elif pair.status.lower() == "paused":
+                        paused_count += 1
                     pair_count += 1
```
nnDarshan: @shtripat as per aravinda, If a particular node is down then the pair info for bricks in the node will not appear as part of get state. which might lead to below logic going wrong. So its better to find the pair count by counting the number of bricks in master volume (because nos of  pairs in geo-rep session = nos of bricks in master volume).

shtripat: Yes. thats a good point. Will change the logic accordingly.

nnDarshan:
The changes look good to me

tendrl/gluster_integration/sds_sync/georep_details.py
```diff
@@ -172,6 +172,10 @@ def aggregate_session_status():
                 )
             except etcd.EtcdKeyNotFound:
                 continue
+            volume = NS.gluster.objects.Volume(vol_id=vol_id).load()
```
r0h4n: use exists() instead of load()

shtripat: Anyway I need volume property brick_count below so thought of loading it.

shtripat:
>> Verified?

Not verified yet.

>> Impact on monitoring-integration and API?

There would be impact in UX I feel as the cluster level georep session is a string value. The correct mapping UI needs to display it in dashboard.

julienlim:
@shtripat Ack - the required changes being proposed makes sense.

shtripat:
@cloudbehl please check this and based on this let me know the changes which are required in monitoring-integration required for grafana dashboard. I will do the needful.

Issue #461, opened 2017-11-02
Alerts which is raised from gluster-integration does not have integration id in message


GowthamShanmugam:
(e.g)

Status of brick: dhcp42-68.lab.eng.blr.redhat.com:/gluster/brick1 under volume volume_test changed from Stopped to Started

need to add integration id in message

GowthamShanmugam:
This issue is fixed, i am closing this

PR #464, opened 2017-11-02
Added integration id in alert message

GowthamShanmugam:
tendrl-bugid: Tendrl/gluster-integration#461

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -172,12 +172,13 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                                 ]
                                 if stored_peer_status != "" and \
                                     current_status != stored_peer_status:
-                                    msg = ("Status of peer: %s "
+                                    msg = ("Status of peer: %s in cluster %s "
                                            "changed from %s to %s") % (
                                                peers[
                                                    'peer%s.primary_hostname' %
                                                    index
                                                ],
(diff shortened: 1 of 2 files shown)


GowthamShanmugam:

GowthamShanmugam:
@shtripat @r0h4n 
In the alert message, while display alert in UI and mail notification we unable to identify the alert belongs to which cluster because message contains no integration id.So in UI also it is difficult to filter the alert using integration id.
so i added integration id also in all status message

shtripat:
Changes look fine.

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -172,12 +172,13 @@ def run(self):
                                 ]
                                 if stored_peer_status != "" and \
                                     current_status != stored_peer_status:
-                                    msg = ("Status of peer: %s "
+                                    msg = ("Status of peer: %s in cluster %s "
```
r0h4n: Is the UI or someone else going to extract the integration_id from msg?

GowthamShanmugam: @r0h4n no, we are not going to extract anything from msg. It just to understand the user easily like this alert belongs to which cluster.

GowthamShanmugam: if volume alert or brick alert came it is difficult to find the alert belongs to which cluster volume and brick, because message contains only volume or brick name.there is no cluster name.

Issue #465, opened 2017-11-06
Gluster-integration sync will fail when ttl delete volume entry from etcd


GowthamShanmugam:
when ttl will delete volume entry in provisioning after this line https://github.com/Tendrl/gluster-integration/blob/master/tendrl/gluster_integration/sds_sync/__init__.py#L236 then gluster integration sync will raise an exception from this loop https://github.com/Tendrl/gluster-integration/blob/master/tendrl/gluster_integration/sds_sync/client_connections.py#L17 and entire gluster-sync is stopped.  

Need to pass volumes only which are not deleted.

GowthamShanmugam:
@shtripat @nnDarshan @r0h4n

GowthamShanmugam:
This issue is fixed, i am closing this

PR #467, opened 2017-11-07
Omitting deleted volumes while collection global cluster details

GowthamShanmugam:
Tendrl-bugid: Tendrl/gluster-integration#465

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -237,7 +237,7 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                     all_volumes = NS.gluster.objects.Volume().load_all() or []
                     volumes = []
                     for volume in all_volumes:
-                        if not eval(volume.deleted):
+                        if not volume.deleted.lower() == "true":
                             volumes.append(volume)
                     cluster_status.sync_cluster_status(volumes)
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@shtripat @r0h4n @nnDarshan please review

GowthamShanmugam:
yes tested

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -234,7 +234,11 @@ def run(self):
                 # Sync cluster global details
                 if "provisioner/%s" % NS.tendrl_context.integration_id \
                     in NS.node_context.tags:
-                    volumes = NS.gluster.objects.Volume().load_all() or []
+                    all_volumes = NS.gluster.objects.Volume().load_all() or []
+                    volumes = []
+                    for volume in all_volumes:
+                        if not eval(volume.deleted):
```
r0h4n: The **eval** function lets a python program run python code within itself. It is a security risk to use eval.

Check for volume.delete.lower() == "true"

GowthamShanmugam: ok i will change this

shtripat: Also only to check the value of `volume.deleted` flag, why we need eval?

GowthamShanmugam:
@r0h4n modifier and verified

Issue #475, opened 2017-11-09
gluster sync fails due to TypeErrror Exception


GowthamShanmugam:
Gluster-integration sync fails due to boolean value is given for lower() function

shtripat:
Closing as the required PR is merged

PR #476, opened 2017-11-09
fix sync error in gluster-integration

GowthamShanmugam:
tendrl-bugid: Tendrl/gluster-integration#475


r0h4n:
Thanks!

Issue #479, opened 2017-11-15
Remove dashbaord creation logic for volume, brick, peer(host) from gluster_integration


GowthamShanmugam:
#Monitoring_integration will keep looking etcd to create a dashboard for volume and brick and host. So the only job is created for delete dashbaord from gluster-integration. Create dashbaord will happen in monitoring-integration sync.

PR #480, opened 2017-11-15
Remove dashbaord creation logic for volume, brick, peer(host) from gl…

GowthamShanmugam:
…uster_integration

tendrl-bugid: Tendrl/gluster-integration#479


GowthamShanmugam:
@rishubhjain @shtripat @r0h4n please review

Issue #481, opened 2017-11-16
Collect inode usage for volumes in etcd


nnDarshan:
Currently inode usage for volume is not stored in etcd. This has to be collected and stored in etcd as volume attribute

PR #482, opened 2017-11-16
Collect inode usage details for volume

nnDarshan:
Currently volume inode usage details are not
collected. This patch takes care of that by
adding attributes to volume object that hold
inode usage info. This is being collected
through libgfapi.

tendrl-bug-id: Tendrl/gluster-integration#481

### DIFF
--- a/tendrl/gluster_integration/objects/definition/gluster.yaml
+++ b/tendrl/gluster_integration/objects/definition/gluster.yaml
@@ -961,6 +961,24 @@ namespace.gluster:
         quorum_status:
           help: "Quorum status"
           type: String
+        usable_capacity:
+          help: "total capacity of the volume"
+          type: int
+        used_capacity:
+          help: "used capacity of the volume"
+          type: int
+        pcnt_used:
+          help: "percentage of space used in volume"
+          type: int
+        total_inode_capacity:
+          help: "total inode"
+          type: int
+        used_inode_capacity:
+          help: "inode used"
+          type: int
+        pcnt_inode_used:
+          help: "inode used percentage"
+          type: int
         profiling_enabled:
           help: Whether profiling is enabled
           type: String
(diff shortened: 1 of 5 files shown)


nnDarshan:
@r0h4n @nthomas-redhat @shtripat @cloudbehl Please review

nnDarshan:
Patch is verified

tendrl/gluster_integration/objects/definition/gluster.yaml
```diff
@@ -961,6 +961,24 @@ namespace.gluster:
         quorum_status:
           help: "Quorum status"
           type: String
+        usable_capacity:
```
r0h4n: `total_capacity` would be better here

tendrl/gluster_integration/objects/definition/gluster.yaml
```diff
@@ -961,6 +961,24 @@ namespace.gluster:
...
+          help: "total capacity of the volume"
+          type: int
+        used_capacity:
+          help: "used capacity of the volume"
+          type: int
+        pcnt_used:
+          help: "percentage of space used in volume"
+          type: int
+        usable_inode_capacity:
```
r0h4n: `total_inode_capacity` would be better

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -33,6 +33,9 @@ def sync_utilization_details(volumes):
             volume.usable_capacity = int(util_det['total'])
             volume.used_capacity = int(util_det['used'])
             volume.pcnt_used = str(util_det['pcnt_used'])
+            volume.usable_inode_capacity = int(util_det['total_inode'])
+            volume.used_inode_capacity = int(util_det['used_inode'])
+            volume.pcnt_inode_used = str(util_det['pcnt_inode_used'])
```
r0h4n: why using str for percentage?

nnDarshan: As of now, obj attrs cannot be of float type. So to preserve the precision str is being used

tendrl/gluster_integration/objects/definition/gluster.yaml
nnDarshan: ack

tendrl/gluster_integration/objects/definition/gluster.yaml
nnDarshan: I would not change it in this PR, this was added as it was missing in defn file. This attr is being used in many places. Better to take later.

Issue #494, opened 2017-11-22
Handle remove brick correctly when subvolume number shifts


nnDarshan:
In cases where subvolume number for a brick shifts after brick deletion, The brick information stored in etcd is entering an inconsistent state.

PR #495, opened 2017-11-22
Handle remove brick correctly when subvolume number shifts

nnDarshan:
As part of remove brick handler we are deleting all the brick
related information, so that next sync takes care of updating
the correct info

tendrl-bug-id: Tendrl/gluster-integration#494

### DIFF
--- a/tendrl/gluster_integration/message/callback.py
+++ b/tendrl/gluster_integration/message/callback.py
@@ -22,6 +22,9 @@ RESOURCE_TYPE_VOLUME = "volume"
 
 
 class Callback(object):
+    def __init__(self):
+        self.sync_interval = NS.config.data.get("sync_interval", 10)
+
     def quorum_lost(self, event):
         context = "quorum|" + event['message']['volume']
         message = "Quorum of volume: {0} is lost in cluster {1}".format(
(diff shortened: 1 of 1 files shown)


tendrl/gluster_integration/message/callback.py
```diff
@@ -19,6 +19,7 @@
 RESOURCE_TYPE_BRICK = "brick"
 RESOURCE_TYPE_PEER = "host"
 RESOURCE_TYPE_VOLUME = "volume"
+SYNC_INTERVAL = NS.config.data.get("sync_interval", 10)
```
GowthamShanmugam: It will give error, NS is not available at this stage

GowthamShanmugam: callback import is happening before NS create https://github.com/Tendrl/gluster-integration/blob/master/tendrl/gluster_integration/manager/__init__.py#L13.


So we can't use NS  here

nnDarshan: Thanks for pointing that.

GowthamShanmugam:
LGTM

nnDarshan:
@r0h4n your changes look good

PR #500, opened 2017-11-23
Fix delete dashboard

GowthamShanmugam:


tendrl/gluster_integration/message/callback.py
```diff
@@ -501,7 +501,10 @@ def volume_delete(self, event):
                                              fqdn,
                                              path
                                          )
-
+                            # fetch brick_path to remove dashboard
+                            brick_full_path = NS._int.wclient.read(
+                                "%s/brick_path" % brick_path
```
r0h4n: put this in a try/except , we dont want issues in one brick stop the entire volume delete

r0h4n:
Verified?

tendrl/gluster_integration/message/callback.py
GowthamShanmugam: i will change it

GowthamShanmugam: It won't fail, it will take next volume when it fails, This code is under try block

GowthamShanmugam:
@r0h4n verified

Issue #514, opened 2017-11-28
Peer reject event should be handled correctly


nnDarshan:
Peer reject event returns peer name with some number appended to it. This is causing issues to our logic which handles duplicates

nnDarshan:
Fix for this issue is merged.Closing this issue

PR #515, opened 2017-11-28
Handle peer reject event correctly

nnDarshan:
peer reject event gives peer name along with a number
appended to it. This is handled correctly by this patch

tendrl-bug-id: Tendrl/gluster-integration#514


nnDarshan:
> Hope both the events are verified with these changes.

peer_reject is verified. But unknown_peer is a very rare event thats hard to simulate(input from atin)

Issue #497, opened 2017-11-22
Rebalance status is not aggregated correctly


nnDarshan:

nnDarshan:
Fix for this issue is merged.Closing this issue

PR #522, opened 2017-11-30
Include arbiter volume for rebalance aggregation

nnDarshan:
* Aggregation was not considering arbiter volume as we had a check for "distribute" in vol type and arbiter volume has "arbiter" as type
* Earlier our understanding was, get state would report rebalance related info for only nodes that were involved in rebalance for a volume. But it reports info about rebalance on all the cluster nodes and if node is not part of rebalance operation it reports the default values (for status its `not stated`) Which messes up the calculation of aggregated status.

This PR rectifies these 2 issues

tendrl-bug-id: Tendrl/gluster-integration#497


nnDarshan:
> I thought we fixed this, could you add more info here or on the issue?

@r0h4n Have updated this PR description. Pls have a look

PR #526, opened 2017-12-01
Set TTL for cluster status and handle it accordingly

shtripat:

### DIFF
--- a/tendrl/gluster_integration/sds_sync/cluster_status.py
+++ b/tendrl/gluster_integration/sds_sync/cluster_status.py
@@ -4,7 +4,7 @@ from tendrl.commons.utils import cmd_utils
 from tendrl.gluster_integration.sds_sync import event_utils
 
 
-def sync_cluster_status(volumes):
+def sync_cluster_status(volumes, sync_ttl):
     # Calculate status based on volumes status
     degraded_count = 0
     is_healthy = True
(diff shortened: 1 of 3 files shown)


tendrl/gluster_integration/sds_sync/cluster_status.py
```diff
@@ -59,7 +59,7 @@ def sync_cluster_status(volumes):
         peer_count=peer_count,
         vol_count=len(volumes),
         volume_up_degraded=degraded_count
-    ).save()
+    ).save(sync_ttl)
```
nnDarshan: pls pass as keyword argument

shtripat: Ah yes. Will do that.

shtripat:
Verified the PR with https://github.com/Tendrl/gluster-integration/pull/527 and cluster goes to `unhealthy` state once all the gluster nodes are brought down. Also the brick status fields vanish after TTL under `clusters/{int-id}/Bricks/all/{node-id}/{brick-name}`.

Also in grafana the cluster is marked as `unhealthy` and volume count is set as all zeros with all the hosts marked as `Down`. Once we get the grafana stale time marking it should ideally look fine in grafana dashboard. @cloudbehl we need the grafana PR now.

tendrl/gluster_integration/sds_sync/cluster_status.py
r0h4n: Are you making this change? @shtripat

shtripat: Ah yes. Looks like didnt push the changes. will do that now

Issue #525, opened 2017-12-01
Add ttl to brick status


nnDarshan:
Add ttl to brick status attribute. Because if all the nodes in the cluster goes down, the status will remain as up and the same info will be displayed in dashboard. To avoid this we set ttl, so thet dashboard can clearly indicate the staleness of the field

nnDarshan:
Fix for this issue is merged. Closing this issue

PR #527, opened 2017-12-01
Add ttl to brick status attribute.

nnDarshan:
Add ttl to brick status attribute. Because if all the
nodes in the cluster goes down, the status will remain
as up and the same info will be displayed in dashboard.
To avoid this we set ttl, so thet dashboard can clearly
indicate the staleness of the field

tendrl-bug-id: Tendrl/gluster-integration#525

### DIFF
--- a/tendrl/gluster_integration/objects/brick/__init__.py
+++ b/tendrl/gluster_integration/objects/brick/__init__.py
@@ -1,4 +1,5 @@
 from tendrl.commons import objects
+from tendrl.commons.utils import etcd_utils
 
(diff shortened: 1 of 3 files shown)


nnDarshan:
verified by stopping gluster-integration and the status field dissappeared for brick

nnDarshan:
@shtripat said he would be taking care of it

shtripat:
taken care of sync stuff as part of https://github.com/Tendrl/gluster-integration/pull/526

PR #530, opened 2017-12-06
Corrected cluster status changes

shtripat:

### DIFF
--- a/tendrl/gluster_integration/objects/global_details/__init__.py
+++ b/tendrl/gluster_integration/objects/global_details/__init__.py
@@ -22,6 +22,7 @@ class GlobalDetails(objects.BaseObject):
         return super(GlobalDetails, self).render()
 
     def save(self, update=True, ttl=None):
+        self.invalidate_hash()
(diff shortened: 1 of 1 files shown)


tendrl/gluster_integration/sds_sync/cluster_status.py
```diff
@@ -54,12 +54,14 @@ def sync_cluster_status(volumes, sync_ttl):
...
-    NS.gluster.objects.GlobalDetails(
+    gd_obj = NS.gluster.objects.GlobalDetails(
         status='healthy' if is_healthy else 'unhealthy',
         peer_count=peer_count,
         vol_count=len(volumes),
         volume_up_degraded=degraded_count
-    ).save(ttl=sync_ttl)
+    )
+    gd_obj.invalidate_hash()
+    gd_obj.save(update=False, ttl=sync_ttl)
```
r0h4n: I would say you simply call self.invalidate_hash in the save() for GlobalDetails,  not here

PR #535, opened 2017-12-07
Changed the logic for sync_ttl for cluster entities

shtripat:


shtripat:
@r0h4n plz review. This solves the issues we discussed around volume and cluster states changing frequently while sync interval.

PR #549, opened 2018-01-11
Raising brick alert when peer is disconnected or node goes down

GowthamShanmugam:
tendrl-bugid: Tendrl/gluster-integration#548

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -32,7 +32,7 @@ from tendrl.gluster_integration.sds_sync import utilization
 RESOURCE_TYPE_BRICK = "brick"
 RESOURCE_TYPE_PEER = "host"
 RESOURCE_TYPE_VOLUME = "volume"
-
+BRICK_STATUS = ["Started", "Stopped"]
 
 class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
 
@@ -157,6 +157,7 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                 if "Peers" in raw_data:
                     index = 1
                     peers = raw_data["Peers"]
+                    disconnected_hosts = []
                     while True:
                         try:
                             peer = NS.gluster.\
@@ -204,6 +205,14 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                                         'Connected'
                                         else 'INFO'
                                     )
+                                    # Disconnected host name to raise brick alert
+                                    if current_status == "Disconnected":
+                                        disconnected_hosts.append(
+                                            peers[
+                                                'peer%s.primary_hostname' %
+                                                index
+                                            ]
+                                        )
                             except etcd.EtcdKeyNotFound:
                                 pass
                             SYNC_TTL += 5
@@ -211,7 +220,12 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                             index += 1
                         except KeyError:
                             break
-
+                    # Raise an alert for bricks when peer disconnected
+                    # or node goes down
+                    for disconnected_host in disconnected_hosts:
+                        brick_status_alert(
+                            disconnected_host
+                        ) 
                 if "Volumes" in raw_data:
                     index = 1
                     volumes = raw_data['Volumes']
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@shtripat @nthomas-redhat @r0h4n please review

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -684,3 +698,80 @@ def sync_volumes(volumes, index, vol_options, sync_ttl):
...
+            NS.tendrl_context.integration_id,
+            hostname
+        )
+    )
+    for brick_info in bricks.leaves:
+        try:
+            brick_path = brick_info.key
+            lock = None
+            lock = etcd.Lock(
+                NS._int.client,
```
GowthamShanmugam: Here read lock is used to avoid the rasing same alert for multiple times by different nodes, the problem is if all the nodes raising the brick alert at the same time then  node_agent in different nodes will raise the same alert.

If we use sleep rather than lock then peer disconnect alert will be raised by different node because it is executed in thread.

GowthamShanmugam: So the read lock will solve the problem

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -684,3 +698,80 @@ def sync_volumes(volumes, index, vol_options, sync_ttl):
...
+                brick_path[1:]
```
shtripat: `brick_path[1:]` just to avoid initial `/` from path ??

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -684,3 +698,80 @@ def sync_volumes(volumes, index, vol_options, sync_ttl):
...
+            lock.acquire(
+                blocking=True,
+                lock_ttl=60
+            )
+            if lock.is_acquired:
+                brick = NS.gluster.objects.Brick(
+                    fqdn=hostname,
+                    brick_dir=brick_path.split("/")[-1]
+                ).load()
+                if brick.status == BRICK_STATUS[0]:
```
shtripat: If this piece of code gets executed only for disconnected hosts, then I feel you can safely say that brick is stopped.

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -684,3 +698,80 @@ def sync_volumes(volumes, index, vol_options, sync_ttl):
...
+                    # raise an alert for brick
+                    msg = ("Status of brick: %s "
+                           "under volume %s in cluster %s chan"
+                           "ged from %s to %s") % (
+                               brick.brick_path,
+                               brick.vol_name,
+                               NS.tendrl_context.integration_id,
+                               BRICK_STATUS[0],
```
shtripat: Also I would suggest to use constants for brick status `Started` and `Stopped` and not an array.

tendrl/gluster_integration/sds_sync/__init__.py
GowthamShanmugam: yes, if / came first then lock is not working, so i removed this

tendrl/gluster_integration/sds_sync/__init__.py
GowthamShanmugam: replace array by constant?

tendrl/gluster_integration/sds_sync/__init__.py
GowthamShanmugam: i will change it

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: yes

tendrl/gluster_integration/sds_sync/__init__.py
GowthamShanmugam: replace this statement by brick_stop constant right?

shtripat: yes

GowthamShanmugam:
@shtripat changes done

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -684,3 +700,80 @@ def sync_volumes(volumes, index, vol_options, sync_ttl):
...
         except KeyError:
             break
+
+
+def brick_status_alert(hostname):
+    # fetching brick details of disconnected node
+    bricks = NS._int.client.read(
+        "clusters/%s/Bricks/all/%s" % (
+            NS.tendrl_context.integration_id,
+            hostname
```
r0h4n: cant you use NS.gluster.objects.Brick.load_all()  ?

GowthamShanmugam: We can i am not sure how it will behave for lock, i will try after i comment

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -207,14 +210,27 @@ def run(self):
                                         'Connected'
                                         else 'INFO'
                                     )
+                                    # Disconnected host name to raise brick alert
+                                    if current_status == "Disconnected":
```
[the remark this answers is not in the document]
GowthamShanmugam: i will change it

GowthamShanmugam:
@r0h4n changes done

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -32,6 +32,8 @@
 RESOURCE_TYPE_BRICK = "brick"
 RESOURCE_TYPE_PEER = "host"
 RESOURCE_TYPE_VOLUME = "volume"
+BRICK_STOPPED = "stopped"
```
r0h4n: On L25, replace with "from tendrl.commons.utils import event_utils"

instead of 
 from tendrl.gluster_integration.sds_sync import event_utils

GowthamShanmugam: i will fix this

GowthamShanmugam:
@r0h4n conflict solved

GowthamShanmugam:
@shtripat @r0h4n please review this

PR #557, opened 2018-02-12
Added Test case for GlusterIntegrationSdsSyncStateThread

Ritsyy:


r0h4n:
Looks like the coverage will decrease due the test, can you fix this?

Ritsyy:
Yes, fixing it, thanks!

Issue #553, opened 2018-02-06
Support for third option for retaining the mixed profiling enabled for volumes


shtripat:
From gluster 3.4 onwards there is going to a flag saying if profiling is enabled for a volume in `get-state` command output.
Backend needs to undergo changes to support this third option while import cluster. If this option selected during import cluster flow, the tendrl should retain the mixed profiling enabled state for the individual volumes and mark the cluster level profiling enabled information as `mixed`.

Also backend should be backward compatible which makes sure, even if we dont have `get-state` returning profile info a volume level, leave the cluster as is and don't mark for individual volumes.

@r0h4n @julienlim @jjkabrown1 @nthomas-redhat comments, suggestions?

julienlim:
@shtripat Ack on backend needing to undergo changes to support Gluster 3.4 get-state support for a profiling flag.

Re: backward compatibility, i.e. < Gluster 3.4, if volume profiling option (radio button in UI) set to "mixed" (leave as-is), I agree that we leave cluster as-is and not make any volume profiling changes.

nthomas-redhat:
Current implementation of enable/disable profiling at cluster level hasdrawbacks and really confusing. If `enable` is selected, profiling will be enabled for all volumes - no issues here. But if user choose to disable the profiling, tendrl is will not disable profiling during the import- rather the volumes with prfiling already on stay as is. This is an inconsitency which needs to be fixed. Also cluster wide profiling status doesn't really make any sense because this is property on a volume level. So this is what I propose:

1. Remove cluster wide profiling status.
2. Profiling status to be added at individual volume level.
3. During import provide three options: a) enable b)disable c)no change. Default will be enabled, perform the specific action based on the user action.
4. Provide an action at cluster level to enable/disable profiling for all volumes in the cluster.
5. Provide an action at volume level to enable/disable profiling for individual volumes.

Let me know what you folks think about this - @shtripat , @r0h4n , @julienlim

julienlim:
@nthomas-redhat @shtripat @r0h4n @gnehapk @a2batic @cloudbehl @mcarrano @Tendrl/tendrl-qe @jjkabrown1

(1) Regarding Removing cluster-wide profiling status, as we allow the option of setting this at a cluster level, we should show the status otherwise, how does user get feedback if they set it at a global level.

(2) +1 on including Profiling status at the volume level, i.e. column added to the Volume list view.  I wasn't sure if we're taking RFE's for this at the moment.  Depends on (5) though.

[...]

PR #559, opened 2018-02-14
Added logic to consider profiling enabled for few volumes

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#553

### DIFF
--- a/tendrl/gluster_integration/objects/definition/gluster.yaml
+++ b/tendrl/gluster_integration/objects/definition/gluster.yaml
@@ -737,7 +737,61 @@ namespace.gluster:
           run: gluster.objects.Volume.atoms.CheckVolumeAvailable
           type: Start
           uuid: 242f6190-9b37-11e6-950d-a24fc0d96765
+        StartProfiling:
+          enabled: true
+          inputs:
+            mandatory:
+              - Volume.vol_id
+              - Volume.volname
+          name: start volume profiling
+          run: gluster.objects.Volume.atoms.StartProfiling
+          type: Start
+          uuid: 242f6190-9b37-11e6-950d-a24fc0d96c05
+          help: Start volume profling
+        StopProfiling:
+          enabled: true
+          inputs:
+            mandatory:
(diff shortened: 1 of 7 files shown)


shtripat:
@r0h4n @nthomas-redhat please review

shtripat:
Verified the changes and if for few volumes profiling is enabled from CLI before import, the individual volumes are marked properly and cluster is marked profiling_enabled as `mixed`.

Need to verify the code with older version of glusterfs where `get-state` doesnt contain the profiling changes.

shtripat:
@r0h4n @nthomas-redhat @a2batic @gnehapk plz review.
Need to add unit tests for additional flows and atoms.

tendrl/gluster_integration/objects/definition/gluster.yaml
```diff
@@ -253,6 +253,12 @@ namespace.gluster:
         tags:
           help: optional parameters
           type: Dict
+        current_job:
```
r0h4n: why are these new attributes added to NativeEvent object? It should be for the Volume object

shtripat: oh thats a mistake. will correct.

tendrl/gluster_integration/objects/definition/gluster.yaml
```diff
@@ -737,7 +737,61 @@ namespace.gluster:
...
+          type: Start
+          uuid: 242f6190-9b37-11e6-950d-a24fc0d96c05
+          help: Start volume profling
+        StopProfiling:
+          enabled: true
+          inputs:
+            mandatory:
+              - Volume.vol_id
+              - Volume.volname
+          name: delete_volume
```
r0h4n: why the name "delete_volume" ?

shtripat: Ctrl C + Ctrl V issue :) will correct.

shtripat:
>> Verified?

Yes

PR #565, opened 2018-02-22
Set first_sync_done once first sync gets over for cluster

shtripat:
Instead of setting cluster.is_managed post sync, now set the field
first_sync_done only for the first time. This makes sure import
cluster flow gets completed only once all the storages nodes report
their first sync done.

tendrl-bug-id: Tendrl/commons#829

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -283,7 +283,13 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                     _cluster = _cluster.load()
                     _cluster.status = ""
                     _cluster.last_sync = str(tendrl_now())
-                    _cluster.is_managed = "yes"
+                    # Mark the first sync done flag
+                    _cnc = NS.tendrl.objects.ClusterNodeContext(
+                        node_id=NS.node_context.node_id
+                    ).load()
+                    if _cnc.first_sync_done in [None, "no"]:
+                        _cnc.first_sync_done = "yes"
+                        _cnc.save()
                     _cluster.save()
                     # Initialize alert count
                     try:


shtripat:
@r0h4n @nthomas-redhat @GowthamShanmugam plz review

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -283,7 +283,24 @@ def run(self):
                     _cluster = _cluster.load()
                     _cluster.status = ""
                     _cluster.last_sync = str(tendrl_now())
-                    _cluster.is_managed = "yes"
+                    # Mark the first sync done flag
+                    try:
+                        NS._int.client.read(
+                            "/clusters/%s/nodes/%s/first_sync_done" %
+                            (
```
r0h4n: Please add this as attribute to object  https://github.com/Tendrl/commons/blob/master/tendrl/commons/objects/cluster_node_context/__init__.py

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -283,7 +283,15 @@ def run(self):
                     _cluster = _cluster.load()
                     _cluster.status = ""
                     _cluster.last_sync = str(tendrl_now())
-                    _cluster.is_managed = "yes"
+                    # Mark the first sync done flag
+                    _cnc = NS.tendrl.objects.ClusterNodeContext(
+                        node_id=NS.node_context.node_id
+                    ).load()
+                    if _cnc.first_sync_done is None or \
```
r0h4n: simplify this   (None and "" are the same)

if _cnc.first_sync_done in [None, "no"] : _cnc.first_sync_done = "yes"

shtripat: done

PR #570, opened 2018-02-27
Corrected issue while setting current job status

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#553

### DIFF
--- a/tendrl/gluster_integration/objects/volume/atoms/start_profiling/__init__.py
+++ b/tendrl/gluster_integration/objects/volume/atoms/start_profiling/__init__.py
@@ -1,3 +1,5 @@
+import time
+
 from tendrl.commons import objects
 from tendrl.commons.objects import AtomExecutionFailedError
 from tendrl.commons.utils import cmd_utils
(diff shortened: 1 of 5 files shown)


shtripat:
@Tendrl/admins please review. This is a regression fix.

tendrl/gluster_integration/objects/volume/atoms/start_profiling/__init__.py
```diff
@@ -25,5 +27,12 @@ def run(self):
                     err
                 )
             )
+        while True:
+            volume = NS.gluster.objects.Volume(
+                vol_id=vol_id
+            ).load()
+            if volume.profiling_enabled == "True":
```
r0h4n: can we start using  "yes" or "no" everywhere boolean?

shtripat: We can I feel. will need to check the dependency at least in this case and change in all references. Will update the PR soon.

PR #572, opened 2018-02-27
Attributes of Tendrl objects should be watchable

r0h4n:
tendrl-bug-id : Tendrl/commons#837


shtripat:
Looks fine. But need a working end to end implementation for at least one object.

Issue #580, opened 2018-03-09
Uncertain what volume profiling event message means and why polling for volume profiling events being shown in UI


julienlim:
In the Tendrl UI (tendrl 1.6.1), some of the event messages shown include:

Value of volume profiling for volume: volume_gama_disperse_4_plus_2x2 of cluster 1bc8cd0e-e24a-493c-ab40-04a121db499e changed from None to no

First question is what is meant by "no" vs. "None" -- isn't it the same thing, or is there some other thing going on?

Second question is I've imported the cluster once, a volume profiling check seems to appear regularly like it's being polled every 3 minutes and generating messages accordingly, and note  there is no related Tasks in the UI, but it shows a job finishing successfully every 3 minutes in the Events List.  Have we decided to expose polling messages in the UI?

Sample event message stream that appears after each 3 minute polling):
```
Job finished successfully (job_id: 6c131bd9-4c2c-409b-b47c-70f8267a2a22)
09 Mar 2018 13:46:36
Value of volume profiling for volume: volume_gama_disperse_4_plus_2x2 of cluster 1bc8cd0e-e24a-493c-ab40-04a121db499e changed from None to no
09 Mar 2018 13:46:12
Value of volume profiling for volume: volume_gama_disperse_4_plus_2x2 of cluster 1bc8cd0e-
e24a-493c-ab40-04a121db499e changed from None to no
09 Mar 2018 13:44:21
```

<img width="1002" alt="screen shot 2018-03-09 at 4 22 53 pm" src="https://user-images.githubusercontent.com/5589919/37230624-2e522708-23b6-11e8-9672-1f02d86c02fe.png">

@r0h4n @shtripat @Tendrl/qe @gnehapk @a2batic

shtripat:
@julienlim this seems an issue. We need to check old value and if it none and first time sync is setting it, it should not raise an event. Also I will check the repeated jobs (I remember @fbalak reported that he sees ConfigureMonitoring jobs getting created and succeeded regularly in tendrl).

I will track the changes under this issue and send fixes.

PR #581, opened 2018-03-10
Check the old value of volume profiling and then raise event

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#580


shtripat:
DONT MERGE. NEED TO VERIFY.

shtripat:
@GowthamShanmugam please apply this in the setup you have and can be verified. Its small thing while volume state changes.

shtripat:
verified

shtripat:
@r0h4n plz review

PR #584, opened 2018-03-13
Volume alert count obj

GowthamShanmugam:


GowthamShanmugam:
Verified

GowthamShanmugam:
This will PR affect UI

GowthamShanmugam:
I have changed volume_alert_counter object attribute name from warn_count to alert_count so UI changes is required @gnehapk

PR #587, opened 2018-03-15
During expansion gluster integration sds sync should be allowed

shtripat:


shtripat:
verified

shtripat:
@r0h4n plz review

PR #595, opened 2018-03-21
Set the cluster status as unhealthy is import failed earlier

shtripat:
tendrl-bug-id: Tendrl/commons#874


shtripat:
DONT MERGE. NEED TO VERIFY.

shtripat:
Verified the changes with import cluster flow. If import fails due to some issues (e.g. improper glusterfs repo so that glusterfs-events install fails), the `cluster.global_details.status` is set as `unhealthy`. In case of successful import cluster the same field is set properly as `healthy`. Works as expected.

PR #597, opened 2018-03-23
Added a check for cluster status as well while saving

shtripat:


shtripat:
verified

PR #602, opened 2018-04-02
Support gluster peer fqdn/ip

r0h4n:
tendrl-bug-id: Tendrl/commons#852


tendrl/gluster_integration/objects/definition/gluster.yaml
```diff
@@ -105,7 +78,7 @@ namespace.gluster:
           help: raw data
           type: String
       enabled: true
-      value: clusters/$TendrlContext.integration_id/raw_map/
+      value: clusters/$TendrlContext.integration_id/nodes/$NodeContext.node_id/raw_map/
```
shtripat: If I understand correctly this for maintaining objects's raw representation. right? If so why under node specific dir?
Can you explain the logic a bit?

r0h4n: Yes. raw map changes per node. Hence saving pee node

shtripat: ok

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -178,17 +178,18 @@ def run(self):
                                     connected=peers['peer%s.connected' % index]
                                 )
                             try:
-                                stored_peer_status = NS._int.client.read(
-                                    "clusters/%s/Peers/%s/connected" % (
-                                        NS.tendrl_context.integration_id,
-                                        peers['peer%s.uuid' % index]
-                                    )
+                                stored_peer_status = etcd_utils.read(
```
shtripat: NS.tendrl.objects.ClusterPeer(...).load() ?

GowthamShanmugam:
LGTM

tendrl/gluster_integration/sds_sync/__init__.py
r0h4n: This block of code will be moved to  GlusterPeer.on_change function soon

PR #608, opened 2018-04-09
Improved alert messages

nthomas-redhat:
tendrl-bug-id:Tendrl/gluster-integration/issues/607

### DIFF
--- a/tendrl/gluster_integration/sds_sync/cluster_status.py
+++ b/tendrl/gluster_integration/sds_sync/cluster_status.py
@@ -51,9 +51,9 @@ def sync_cluster_status(volumes, sync_ttl):
     old_status = cluster_gd.status or 'unhealthy'
     curr_status = 'healthy' if is_healthy else 'unhealthy'
     if curr_status != old_status:
-        msg = ("%s is %s"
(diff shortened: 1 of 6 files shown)


tendrl/gluster_integration/sds_sync/cluster_status.py
```diff
@@ -51,10 +51,9 @@ def sync_cluster_status(volumes, sync_ttl):
     old_status = cluster_gd.status or 'unhealthy'
     curr_status = 'healthy' if is_healthy else 'unhealthy'
     if curr_status != old_status:
-        msg = ("Health status of cluster: %s "
-               "changed from %s to %s") % (
+        msg = ("%s is %s"
```
r0h4n: This msg and some others  have very little context.
eg: "Cluster :cluster_id is :curr_status" , this one with added context

I suggest we atleast prefix object types like "Cluster ", "Node", "Volume" , "Brick"  before we use integration_id, node_id, volume name or brick/node fqdn

Currently most msgs are of type, "some_uuid  in some_other_uuid is down" , which is very difficult to understand without prefixing the resource/object type

@nthomas-redhat @julienlim

Issue #609, opened 2018-04-11
Brick details are not populated when peer hostname and brick hostname not match


GowthamShanmugam:

GowthamShanmugam:
This issue is fixed, i am closing this

PR #610, opened 2018-04-11
Populating brick details correctly when peer hostname and brick hostname not match

GowthamShanmugam:
tendrl-bug-id: Tendrl/gluster-integration#609

Signed-off-by: Gowthamshanmugasundaram <gshanmug@redhat.com>


GowthamShanmugam:
@r0h4n verified

Issue #629, opened 2018-05-08
Gluster native event webhook fails sometimes


r0h4n:
Gluster sends frequent events to tendrl's registered webhook (callback) [0] , it is observed the tendrl webhook fails to accept the http POST request sent by `glustereventsd` [1]

[1] https://pastebin.com/xQrwSDFt

PR #630, opened 2018-05-10
Use cherrypy for gluster native events webhook

r0h4n:
tendrl-bug-id: Tendrl/gluster-integration#629


r0h4n:
@nthomas-redhat please review

GowthamShanmugam:
small correction, otherwise looks good

Issue #631, opened 2018-05-14
Log the errors as task info if volume profile set fails for a volume with profiling enabled already


shtripat:
Currently these are logged as errors in task messages. Rather these should be logged as info messages as they are not errors actually.

shtripat:
Closing as required PRs are merged

PR #633, opened 2018-05-14
Changed to log the message as info if profile flag setting errors out

shtripat:
In case profiling is already enabled for a volume, enable command
fails for the same. In the flow these are logged as errors in task.
Changed the logic to log the messages and info as its actually not
a failure.

tendrl-bug-id: Tendrl/gluster-integration#631
Bugzilla-id: https://bugzilla.redhat.com/1577228


shtripat:
Verified the changes and the message is logged as info if profiling enabled for few volumes and cluster level we mark to enable profiling for all the volumes.

Issue #636, opened 2018-05-15
Dont set cluster.profiling_enabled_state if volume.profiling_enabled flag values are Unknown


shtripat:
This happens if get-state from gluster doesn't report volume specific profiling enabled value. In this case volume level values for profiling_enabled flag would be set as Unknown and so at cluster level we should not set the value of cluster.profiling_enabled_state. Later if user enables/disables profiling from UI the respective value should be set for cluster volume still would report as Unknown because get-state doesnt support the details.

This whole thing is required only for a glusterfs version which doesn't report profiling information at individual volume level.

shtripat:
Closing as required PR is merged

PR #637, opened 2018-05-15
Dont set cluster level profiling flag if volume level flags Unknown

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#636
Bugzilla-id: https://bugzilla.redhat.com/1578333

### DIFF
--- a/tendrl/gluster_integration/flows/enable_disable_volume_profiling/__init__.py
+++ b/tendrl/gluster_integration/flows/enable_disable_volume_profiling/__init__.py
@@ -99,5 +99,6 @@ class EnableDisableVolumeProfiling(flows.BaseFlow):
             'job_name': self.__class__.__name__,
             'job_id': self.job_id
         }
+        _cluster.volume_profiling_state = "%sd" % action
         _cluster.save()
         return True
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -386,9 +386,17 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                     )
                 cluster.volume_profiling_state = "disabled"
         profiling_enabled_count = 0
+        profiling_unknown_count = 0
         for volume in volumes:
             if volume.profiling_enabled == "yes":
                 profiling_enabled_count += 1
+            if volume.profiling_enabled in [None, ""]:
+                profiling_unknown_count += 1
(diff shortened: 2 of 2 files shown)


shtripat:
we need a fix for https://github.com/Tendrl/ui/issues/959 as well for this PR to work properly.
@gnehapk ^^^

shtripat:
With these changes behaviour would be like below

1. In case `gluster get-state` command returns volume specific profiling information
- in UI the volumes would show profiling enabled or disabled
- User would be able to enable/disable profiling at volume level from UI
- cluster level value for profiling state would be calculated based on individual volume's profiling values (all volumes enabled = cluster enabled, all volumes disabled = cluster enabled, few volumes enabled = cluster mixed)
- cluster level option for enabling/disabling profiling for all the volumes of the cluster and individual volumes in UI would show the values accordingly yes enabled or disabled

2. In case `gluster get-state` command doesnt return volume specific profiling information
- in UI the volumess would show profiling enabled value as `Unknown` (as there is no information available for volumes from get-state)
- User wont be able to enable/disable profiling for individual volumes
- cluster level value for profiling enabled flag would be set as None initially after import cluster in tendrl
- User can enable / disable profiling for all the volumes at cluster level. This would set the flag accordingly for underlying volumes in cluster. But UI wont be able to show the same for individual volumes (as get-state doesnt report the volume specific profiling information). At cluster level profiling flag would be set accordingly once enable / disable executed from UI

@julienlim @nthomas-redhat @r0h4n comments? suggestions?
This is what was discussed earlier but there were small issues with existing code so these fixes required.

julienlim:
@shtripat @nthomas-redhat @r0h4n @shirshendu @gnehapk 

In looking at https://github.com/Tendrl/gluster-integration/pull/637#issuecomment-389205316, it looks overall fine except for the following:

For #1 (In case gluster get-state command returns volume specific profiling information), for the cluster level value for profiling state, if all volumes disabled, wouldn't that be cluster disabled (vs. "all volumes disabled = cluster enabled")?

shtripat:
@julienlim ah thats a mistake. I will correct the comment above. Thanks for pointing it out.

julienlim:
+1 LGTM

shtripat:
Verified. @gnehapk I understand you have verified this with UI changes as well. Can you confirm?

gnehapk:
@shtripat Yes. Corresponding PR for it - https://github.com/Tendrl/ui/pull/960.

PR #651, opened 2018-05-28
When child job fails in provisioner node then other child job are not responding

GowthamShanmugam:
tendrl-bug-id: Tendrl/gluster-integration/issues/650
bugzilla: 1570048

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -238,10 +238,11 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                             NS.tendrl_context.integration_id,
                             vol_id=vol_id
                         ).load()
-                        dest = dict(volume.options)
-                        dest.update(dict1)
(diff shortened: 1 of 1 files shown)


tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -238,10 +238,11 @@ def run(self):
                             NS.tendrl_context.integration_id,
                             vol_id=vol_id
                         ).load()
-                        dest = dict(volume.options)
```
GowthamShanmugam: volume option details are only populated by provisioning node if provisioning node child job fails then volume.option is None so dict(volume.options) will exception so below this we are actually doing "first_sync" is yes because of this exception it won't happen so child job stuck and unmanage also fails.

GowthamShanmugam:

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: Ok. Good catch of a corner catch. Thanks for fixing this.

Issue #659, opened 2018-06-14
After node up peer alert is not replaced by clearing alert


GowthamShanmugam:
peer disconnected alert always remains same after node up

PR #660, opened 2018-06-14
After node up peer alert is not replaced by clearing alert

GowthamShanmugam:
tendrl-bug-id: Tendrl/gluster-integration#659
bugzilla: 1519201

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -150,8 +150,17 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                                     connected=peers['peer%s.connected' % index]
                                 )
                             try:
+                                stored_peer_status = None
+                                # find peer detail using hostname
+                                ip = socket.gethostbyname(
+                                    peers['peer%s.primary_hostname' % index]
+                                )
+                                node_id = etcd_utils.read(
+                                    "/indexes/ip/%s" % ip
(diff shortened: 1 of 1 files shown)


tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -150,8 +150,17 @@ def run(self):
...
+                                # find peer detail using hostname
+                                ip  = socket.gethostbyname(
+                                    peers['peer%s.primary_hostname' % index]
+                                )
+                                node_id = etcd_utils.read(
+                                    "/indexes/ip/%s" % ip
+                                ).value
                                 stored_peer = NS.tendrl.objects.GlusterPeer(
-                                    peer_uuid=peers['peer%s.uuid' % index]
+                                    peer_uuid=peers['peer%s.uuid' % index],
```
GowthamShanmugam: peer status always should check to form actual nodes peer directory, Because of a problem with the current approach is then node who is raising peer disconnected alert is only updating itself peer is a disconnect. But after few minutes it is deleted by TTL. So there is no meaning of marking the peer is down. So when no up there is no difference between old status and the new one. Because all other nodes have a value connected only for that particular peer. 

In this logic, I have modified the code like peer status fetching and updating happens actual nodes peer directory. So all nodes always look the same place.

GowthamShanmugam:
@shtripat verified, it works fine

shtripat:
LGTM. Rebase please

PR #661, opened 2018-06-15
Unit tests for sds_sync module

jmolmo:
sds_sync module implementation 
---------------------------------------

In my opinion there is needed a refactorization of the code in this module. Basically the problem is the "extension" of the implementation in each of the methods/functions.
It seems that all the execution units can be easily simplified "breaking" them in small auxiliary functions. 
This will give us several advantages:
1. More easy to read/maintain the code
2. Unit test more easy to do, and a real possibility of test diferent scenes in the unit tests, not only code coverage is important.
3. More easy to extend/add functionality
4. More difficult to repeat things already done/checked in the same function

Method GlusterIntegrationSdsSyncStateThread.Run()
--------------------------------------------------

Cluster state and cluster node context are loaded several times in different parts of the method. 
I think that even in the case that the situation/state of the cluster change while this method is being executed is not good to load again this information.
IMHO, this can lead to inconsistences. I think that it would be safer to load the information at the beginning and use this information in the rest of the method. Periodic syncs will give us the evolution of the state, but refreshing the status information while we are executing the "run" method seems to be dangerous.

I haven't got a good knowledgment of the whole system, and i haven't work with gluster clusters, also not time enough to solve this lacks, so it is possible that i be wrong in my changes, in any case, this is nothing that cannot be solved with a good code review!, and i really think that this changes will improve the code.

Function check_volumes()
-------------------------

This function has been extracted (refactorization) from the GlusterIntegrationSdsSyncStateThread.Run(). The unit test shows that each time a new option is found, it is saved ... probably it can be modified to save only in one attemp all the options.

Another thing i do not unsderstand is why if there is no options in the volume loaded, the options discovered (gluster cmd details) are not included in the volume.

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -41,10 +41,10 @@ def cluster_not_ready(cluster):
     @cluster: A Tendrl Cluster object
     @retuns boolean indicating if the cluster is ready or not
     """
-    status_not_ready = (cluster.status == "importing" and
-                        cluster.current_job['status'] == 'failed') or \
+    status_not_ready = ((cluster.status == "importing" and
+                         cluster.current_job['status'] == 'failed') or \
                         cluster.status == "unmanaging" or \
-                        cluster.status == "set_volume_profiling"
+                        cluster.status == "set_volume_profiling")
 
     return status_not_ready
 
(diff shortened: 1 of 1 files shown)


r0h4n:
@GowthamShanmugam  @anmolsachan please review this 

@shtripat  The changes seem interesting, please review them

shtripat:
Somehow even I am not able to see the comment. Anyway overall code changes look fine.

jmolmo:
Thx!! No problem

:-)

PR #663, opened 2018-06-15
Sometimes volume alert raise with name None

GowthamShanmugam:
tendrl-bug-id: Tendrl/gluster-integration#662
bugzilla: 1588440

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -875,7 +875,9 @@ def get_volume_alert_counts():
         NS.tendrl_context.integration_id
     ).load_all()
     for volume in volumes:
-        alert_counts[volume.name] = {'vol_id': volume.vol_id,
-                                     'alert_count': 0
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -253,10 +253,11 @@ def run(self):
                     ).load_all() or []
                     volumes = []
                     for volume in all_volumes:
-                        if not str(volume.deleted).lower() == "true" or \
-                            volume.current_job.get('status', '') \
-                            in ['', 'finished', 'failed']:
-                            volumes.append(volume)
+                        if volume.exists():
```
shtripat: all_volumes we are getting for etcd only. How come few of them doesnt exist and why we need this check?

GowthamShanmugam: that's correct, I think I misunderstood the logic here, i will remove this

PR #664, opened 2018-06-15
Creating native message handler object after integration_id found

GowthamShanmugam:
Inside message handler constructor we are finding cluster_name
using integration_id, if it before called integration_id found then
it will raise alert with None

tendrl-bug-id: Tendrl/gluster-integration#662
bugzilla: 1588440


GowthamShanmugam:
https://github.com/Tendrl/gluster-integration/blob/master/tendrl/gluster_integration/message/callback.py#L35

GowthamShanmugam:

Issue #667, opened 2018-06-27
Volume alert raised with volume name None


GowthamShanmugam:
It is one more corner case problem, the problem is we down all storage nodes and after volume are deleted by TTL then we are starting all the nodes, what is happening all storage nodes are pushing volume_rebalance data into etcd before we decide provisioner. now volume entry present with rebalance directory but there are no details about volume because volume details are populated by provisioner only. but provisioner is decided in-between of execution then global details are executed before volume details populate. so alert raised with name None.

GowthamShanmugam:
This issue is fixed

PR #668, opened 2018-06-27
Volume alert raised with volume name None

GowthamShanmugam:
It is one more corner case problem, the problem is we
down all storage nodes and after volume are deleted by
TTL then we are starting all the nodes, what is happening
all storage nodes are pushing volume_rebalance data into
etcd before we decide provisioner. now volume entry present
with rebalance directory but there are no details about
volume because volume details are populated by provisioner
only. but provisioner is decided in-between of execution
then global details are executed before volume details
populate. so alert raised with name None.

tendrl-bug-id: Tendrl/gluster-integration#667
bugzilla: 1595295

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -282,9 +282,11 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                     ).load_all() or []
                     volumes = []
                     for volume in all_volumes:
-                        if not str(volume.deleted).lower() == "true" or \
+                        if not str(volume.deleted).lower() == "true" and \
                             volume.current_job.get('status', '') \
-                            in ['', 'finished', 'failed']:
+                            in ['', 'finished', 'failed'] and \
+                            volume.vol_id not in [None, ''] and \
+                            volume.name not in [None, '']:
                             # only for first sync refresh volume TTL
                             # It will increase TTL based on no.of volumes
                             if _cnc.first_sync_done in [None, "no", ""]:
@@ -292,7 +294,7 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                                     volume.value,
                                     SYNC_TTL + VOLUME_TTL
                                 )
-                                volumes.append(volume)
+                            volumes.append(volume)
                     cluster_status.sync_cluster_status(
                         volumes, SYNC_TTL + VOLUME_TTL
                     )


tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -266,9 +266,11 @@ def run(self):
                     ).load_all() or []
                     volumes = []
                     for volume in all_volumes:
-                        if not str(volume.deleted).lower() == "true" or \
+                        if not str(volume.deleted).lower() == "true" and \
                             volume.current_job.get('status', '') \
```
GowthamShanmugam: it should be and not or  @shtripat suggestion please

GowthamShanmugam:

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: I feel it should be as below

```
if not str(volume.deleted).lower() == 'true' or (
    volume.current_job.get('status', '') in ['', 'finished', 'failed'] and
    volume.vol_id not in [None, ''] and
    volume.name not in [None, '']
):
```

What you say?

GowthamShanmugam: if a volume is deleted and other condition or true then it will proceed with deleted volume right?

GowthamShanmugam: i feel all condition or required, or will leads to proceed with deled volume also

shtripat: eventually if you remove the bracket after or the condition becomes like below

```
if not str(volume.deleted).lower() == 'true' or 
    volume.current_job.get('status', '') in ['', 'finished', 'failed'] or
    volume.vol_id not in [None, ''] or
    volume.name not in [None, '']:
```

So we both are saying same thing ideally :)

GowthamShanmugam: sorry i mean or is not required, i wrongly committed, and is required

shtripat:
Would it solve 1596820 as well?

nthomas-redhat:
@GowthamShanmugam

nthomas-redhat:
@GowthamShanmugam , Can we also make sure to not pouplate global details and re-bablace data if the volume details are not populated?

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -266,9 +266,11 @@ def run(self):
...
-                            in ['', 'finished', 'failed']:
+                            in ['', 'finished', 'failed'] and \
+                            volume.vol_id not in [None, ''] and \
+                            volume.name not in [None, '']:
                             volumes.append(volume)
```
GowthamShanmugam: @nthomas-redhat if volume details are not populated and volume option details are populated then volume object will be empty, the check volume.vol_id not in [None, ''] and volume.name not in [None, ''] will takecare of omitting the volume empty volume obj.

nthomas-redhat: ack

nthomas-redhat:
@GowthamShanmugam , resolve the merge conflicts

Issue #669, opened 2018-06-29
Improve gluster integration sync execution performance 


GowthamShanmugam:
Gluster integration sync takes a lot of time to finish one cycle, The problem which we faced with this issue was:

   1. import flow takes a lot of times to finish
   2. When a number of volumes are high then volumes are deleted by TTL before sync complete

GowthamShanmugam:
This issue is fixed

PR #675, opened 2018-07-04
Adjusting volume TTL value

GowthamShanmugam:
tendrl-bug-id: Tendrl/gluster-integration#669
bugzilla: 1596862

### DIFF
--- a/tendrl/gluster_integration/message/callback.py
+++ b/tendrl/gluster_integration/message/callback.py
@@ -489,8 +489,7 @@ class Callback(object):
             if fetched_volume.name == event['message']['name']:
                 fetched_volume.deleted = True
                 fetched_volume.deleted_at = time_utils.now()
-                sync_ttl = int(NS.config.data.get("sync_interval", 10))
-                fetched_volume.save(ttl=sync_ttl)
+                fetched_volume.save()
                 try:
                     sub_volumes = etcd_utils.read(
                         "/clusters/{0}/Volumes/{1}/Bricks".format(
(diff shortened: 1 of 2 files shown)


tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -246,9 +246,9 @@ def run(self):
                         except KeyError:
                             global VOLUME_TTL
                             # from second sync volume ttl is
-                            # SYNC_TTL + (no.volumes) * 30
+                            # SYNC_TTL + (no.volumes) * 30 + 200
```
GowthamShanmugam: I worried when only one volume with large number bricks then  SYNC_TTL + (no.volumes) * 30 not enough. This change is only to handle one or two volume with a large number of brick in the same node.

tendrl/gluster_integration/message/callback.py
```diff
@@ -489,8 +489,7 @@ def volume_delete(self, event):
             if fetched_volume.name == event['message']['name']:
                 fetched_volume.deleted = True
                 fetched_volume.deleted_at = time_utils.now()
-                sync_ttl = int(NS.config.data.get("sync_interval", 10))
-                fetched_volume.save(ttl=sync_ttl)
+                fetched_volume.save()
```
GowthamShanmugam: Let the volume deleted by TTL which assigns from sync thread, because if a cluster is big and volume is very high then assigning 60 sec of default sync interval to delete is not good, because it may delete volume in between of sync.

GowthamShanmugam:

GowthamShanmugam:
Now volume TTL is based on no.of volumes and number of bricks in a cluster

PR #680, opened 2018-07-11
Invoke volume utilization calculation as subprocess

shtripat:
There seems to be some meory leaks in gfapi and due to that if we
make in-memory calls to gfapi, the memory utilization on the nodes
keeps increasing slowly. If we make subprocess call this increase would
be eventual and no overall increase happens to the memory utilization
of the node.

tendrl-bug-id: Tendrl/gluster-integration#681
bugzilla: 1599987

### DIFF
--- a/tendrl/gluster_integration/sds_sync/utilization.py
+++ b/tendrl/gluster_integration/sds_sync/utilization.py
@@ -1,11 +1,15 @@
+import json
+import os
+import subprocess
+
 from tendrl.commons.utils import log_utils as logger
-from tendrl.gluster_integration.sds_sync.vol_utilization import \
-    showVolumeUtilization
 
 
 def sync_utilization_details(volumes):
     cluster_used_capacity = 0
     cluster_usable_capacity = 0
+    volnames = ''
+    voldict = {}
     for volume in volumes:
         if volume.status != "Started":
             logger.log(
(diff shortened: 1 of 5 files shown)


GowthamShanmugam:
this will solve issue: 1599985 also

tendrl/gluster_integration/sds_sync/vol_utilization.py
```diff
@@ -18,42 +19,55 @@ def computeVolumeStats(data):
...
+    volumes = vnames.split(',')
+    ret_dict = {}
+    for volume in volumes:
+        if volume != "":
+            try:
+                data = gfapi.getVolumeStatvfs(volume)
+            except gfapi.GlusterLibgfapiException:
+                sys.stderr.write("CRITICAL: Failed to get the "
+                                 "Volume Utilization Data\n")
+                sys.exit(-1)
```
GowthamShanmugam: I thought we can continue with other volumes if anyone volume is failed here, but i feel if we remove sys.exit(-1) again it will lead to memory leak problem. so this approach is fine.

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -17,28 +21,39 @@ def sync_utilization_details(volumes):
...
-            volume.used_capacity = int(util_det['used'])
-            volume.pcnt_used = str(util_det['pcnt_used'])
-            volume.total_inode_capacity = int(util_det['total_inode'])
-            volume.used_inode_capacity = int(util_det['used_inode'])
-            volume.pcnt_inode_used = str(util_det['pcnt_inode_used'])
+        volnames += volume.name + ','
+        voldict[volume.name] = volume
+    cmd = subprocess.Popen(
+        "tendrl-gluster-vol-utilization %s" % volnames,
+        shell=True,
```
GowthamShanmugam: please comment before this subprocess call like to avoid memory leak we are doing CLI call. Otherwise again someone will change this logic in the future.

GowthamShanmugam:
LGTM

tendrl/gluster_integration/sds_sync/vol_utilization.py
shtripat: Ack. Next time anyway it will try to get details again.

tendrl/gluster_integration/sds_sync/utilization.py
shtripat: Ack :)

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -17,28 +21,55 @@ def sync_utilization_details(volumes):
...
             )
+    except(KeyError, TypeError, ValueError):
+        logger.log(
+            "error",
+            NS.publisher_id,
+            {
+                "message": "Error getting utilization of "
+                "volumes %s . err: %s" % (volnames, out)
+            }
+        )
```
GowthamShanmugam: @shtripat please initialize out='' or out= None after line 39.

shtripat: Ack

Issue #684, opened 2018-07-20
Fix the geo replication session reference in sds_sync


shtripat:
There is no field with name `key`. We need to refer the field value for the session_id for further calculation of pairs of the geo replication session.

PR #685, opened 2018-07-20
Fixed GeoReplicationSession field reference

shtripat:
There is no field with name `key`. We need to refer the field
value for the session_id for further calculation of pairs of
the geo replication session. Corrected the same as part this
change.

tendrl-bug-id: Tendrl/gluster-integration#684
bugzilla: 1603175

### DIFF
--- a/tendrl/gluster_integration/sds_sync/georep_details.py
+++ b/tendrl/gluster_integration/sds_sync/georep_details.py
@@ -1,5 +1,6 @@
 import etcd
 
+from tendrl.commons.utils import etcd_utils
 from tendrl.commons.utils import event_utils
 from tendrl.gluster_integration.objects.geo_replication_session\
     import GeoReplicationSessionStatus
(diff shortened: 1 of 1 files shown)


shtripat:
@GowthamShanmugam @nthomas-redhat please review

tendrl/gluster_integration/sds_sync/georep_details.py
```diff
@@ -187,7 +187,7 @@ def aggregate_session_status():
                 sessions = []
             for session in sessions:
                 session_status = None
-                session_id = session.key.split("GeoRepSessions/")[-1]
+                session_id = session.value.split("GeoRepSessions/")[-1]
                 pairs = NS.gluster.objects.GeoReplicationPair(
```
GowthamShanmugam: can we use session.session_id here?

shtripat: I checked but the field is not populated somehow after load_all

tendrl/gluster_integration/sds_sync/georep_details.py
```diff
@@ -177,20 +178,21 @@ def aggregate_session_status():
             vol_id = volume.vol_id
             sessions = None
             try:
-                sessions = NS.tendrl.objects.GeoReplicationSession(
-                    vol_id=volume.vol_id
-                ).load_all()
+                sessions = etcd_utils.read(
```
nthomas-redhat: Why we are changing this? we can do a load on the object itself right?

GowthamShanmugam: GeoReplicationSession object directry present inside /cluster/{cid}/Volumes/{vid}/GeoRepSessions also in same directry we have GeoReplicationPair (/cluster/{cid}/Volumes/{vid}/GeoRepSessions/{geo_rep_id}/GeoReplicationPair). So we are still not tested to use load all with directory which have child directory. 

if /cluster/{cid}/Volumes/{vid}/GeoRepSessions/{geo_rep_id} does not have any child directory then load_all with work without any problem. But directory with subdirectory still we need to test.

So I and @shtripat decided to use `ectd.utils.read`  to find geo_rep_session ids and using ids we are loading GeoReplicationPair because GeoReplicationPair does not have any subdirectory.

shtripat: @nthomas-redhat so by this time GeoReplicationSession objects are not yet written and we are writing here. To get the session_id we are doing a raw read from path written by GeoReplicationPair objects and then forming the session object down and writing.

nthomas-redhat: ok

nthomas-redhat:
@shtripat , Fix UT failures

Issue #690, opened 2018-08-02
Fix cluster object race condition issue


GowthamShanmugam:

GowthamShanmugam:
This issue is fixed

PR #691, opened 2018-08-02
Fix cluster object race condition issue

GowthamShanmugam:
tendrl-bug-id: Tendrl/gluster-integration#690
bugzilla: 1596655


GowthamShanmugam:
@shtripat @nthomas-redhat please review

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -445,14 +445,16 @@ def _enable_disable_volume_profiling(self):
             if volume.profiling_enabled in [None, ""]:
                 profiling_unknown_count += 1
         if profiling_unknown_count == len(volumes):
-            cluster.save()
             return
```
GowthamShanmugam: Tere is no change is a cluster object from above statements, storing this without loading will leads to race condition, instead of keeping this i am remving this.

Tested

Issue #694, opened 2018-08-16
Svc connected and disconnected alerts are raised only one node, other nodes alerts are not displayed in UI


GowthamShanmugam:

GowthamShanmugam:
This issue is fixed

PR #695, opened 2018-08-16
Raising svc_connected and svc_disconnected for all nodes in a cluster

GowthamShanmugam:
tendrl-bug-id: Tendrl/gluster-integration#694
bugzilla: 1616215, 1616208

### DIFF
--- a/tendrl/gluster_integration/message/callback.py
+++ b/tendrl/gluster_integration/message/callback.py
@@ -69,15 +69,20 @@ class Callback(object):
         native_event.save()
 
     def svc_connected(self, event):
-        context = "svc_connection|" + event['message']['svc_name']
+        context = "svc_connection|" + "%s_%s" % (
+            event['message']['svc_name'],
+            NS.node_context.fqdn
+        )
         volname = event['message'].get('volume', '')
         if volname:
             context += volname
 
-        message = "Service: {0} is connected in cluster {1}".format(
-            event['message']['svc_name'],
-            self.cluster_short_name
-        )
+        message = "Service: {0} is connected on node {1} of " \
+            "cluster {2}".format(
+                event['message']['svc_name'],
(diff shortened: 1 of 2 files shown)


GowthamShanmugam:
@shtripat @nthomas-redhat please review

tendrl/gluster_integration/message/callback.py
```diff
@@ -69,15 +69,20 @@ def quorum_regained(self, event):
         native_event.save()
 
     def svc_connected(self, event):
-        context = "svc_connection|" + event['message']['svc_name']
+        context = "svc_connection|" + "%s_%s" % (
+            event['message']['svc_name'],
+            NS.node_context.fqdn
+        )
```
shtripat: If I understand correctly the context in this scenario would be of form `svc_connection|glusterd_{node-fqdn}` and if the volume name present in message, the context would be `svc_connection|glusterd_{node-fqdn}{volume-name}`.
The question is how are we processing this context? Some comment here would be helpful if you explain how the context would be tackled.

GowthamShanmugam: All callback functions will save the native events into etcd using  `NS.gluster.objects.NativeEvents` object. The provisioner node will call process event function in each gluster-integration sync https://github.com/Tendrl/gluster-integration/blob/38d5d66c1d0f2982edbc2702a9d87e1d20df7864/tendrl/gluster_integration/sds_sync/__init__.py#L263 to process the native events.

Then process event function https://github.com/Tendrl/gluster-integration/blob/38d5d66c1d0f2982edbc2702a9d87e1d20df7864/tendrl/gluster_integration/message/process_events.py#L9 will load all native objects and push that object into the message socket as alerts.

Context is used as `plugin_instance` to a different one alert from another.

anmolsachan:
LGTM

PR #701, opened 2018-10-11
Moved profiling post to import completion

anmolsachan:
tendrl-bug-id: Tendrl/gluster-integration/issues/702
bugzilla: 1638314
Signed-off-by: Anmol Sachan anmol13694@gmail.com

### DIFF
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -328,7 +328,7 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                 # check and enable volume profiling
                 if "provisioner/%s" % NS.tendrl_context.integration_id in \
                     NS.node_context.tags:
-                    self._enable_disable_volume_profiling()
+                    self._update_cluster_volume_profiling_state()
 
                 _cluster = NS.tendrl.objects.Cluster(
                     integration_id=NS.tendrl_context.integration_id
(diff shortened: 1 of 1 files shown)


tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -386,54 +386,7 @@ def _enable_disable_volume_profiling(self):
         # should be done only once while first sync. Later the volume
         # level volume_profiling_state should be set based on individual
         # volume level values
-        _cnc = NS.tendrl.objects.ClusterNodeContext(
```
shtripat: I feel the above comment needs a change to explain that the profile enable/disable would be invoked as a separate flow now.

shtripat: Also the function name should change as its no more enabling/disabling profiling for the volumes. Rather its setting the cluster level value for profiling. Right?

anmolsachan: right, Will make the required change.

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -375,65 +375,17 @@ def run(self):
             {"message": "%s complete" % self.__class__.__name__}
         )
 
-    def _enable_disable_volume_profiling(self):
+    def _update_cluster_volume_profiling_state(self):
         cluster = NS.tendrl.objects.Cluster(
             integration_id=NS.tendrl_context.integration_id
         ).load()
```
GowthamShanmugam: replace this entire cluster object load by a variable volume_profiling_state = None

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -454,8 +406,9 @@ def _enable_disable_volume_profiling(self):
             cluster.volume_profiling_state = "mixed"
         volume_profiling_state = cluster.volume_profiling_state
         cluster = cluster.load()
-        cluster.volume_profiling_state = volume_profiling_state
-        cluster.save()
+        if cluster.volume_profiling_state != volume_profiling_state:
```
GowthamShanmugam: Replace all   cluster.volume_profiling_state = "disabled", cluster.volume_profiling_state = "enabled", cluster.volume_profiling_state = "mixed" by volume_profiling_state = "disabled", volume_profiling_state = "enabled", volume_profiling_state = "mixed"

and modify the if condition like (if cluster.volume_profiling_state != volume_profiling_state and volume_profiling_state != None)

loading cluster object once is enough

shtripat: I agree to this. First calculate the value of new `volume_profiling_state` and at last load the cluster object and update the field.

anmolsachan:
Closing this as a better way was found. Decided to keep profiling in the beginning of the import , with the flow

Issue #703, opened 2018-10-11
Remove blivet dependency from gluster-integration module


GowthamShanmugam:
When a number of bricks are high in a machine then blivet takes huge time to form a device tree. We are using python-blivet in a gluster-integration to fetch device details using mount points When a number of bricks high in the same node then it took a lot of times to collect all devices details and form a device tree. the blivet is collecting very high-level device details but we are using only a few basic device details from that. We have to replace blivet dependency with some other tools to reduce sync time.

GowthamShanmugam:
This issue is fixed

PR #704, opened 2018-10-11
Removing python-blivet dependency

GowthamShanmugam:
bugzilla: 1637977
tendrl-bug-id: Tendrl/gluster-integration#703


GowthamShanmugam:
@shtripat @anmolsachan @nthomas-redhat @cloudbehl please review

tendrl/gluster_integration/sds_sync/brick_device_details.py
```diff
@@ -36,41 +36,16 @@ def update_brick_device_details(brick_name, brick_path, devicetree, sync_ttl):
...
-    brick.pv = pvs
-    brick.size = size
-    brick.save(ttl=sync_ttl)
+    if mount_point in devicetree.keys():
+        brick = NS.tendrl.objects.GlusterBrick(
+            NS.tendrl_context.integration_id,
+            NS.node_context.fqdn,
+            brick_dir=brick_name.split(":_")[-1]
+        ).load()
+        brick.devices = devicetree[mount_point].get("disks", [])
```
shtripat: So what I understand from this is that we need only brick's devices, partitions, size and mount_path. Correct?
And thats what we are getting here from lsblk and populating in brick object.

All other details like lv, vg, pool and pv we dont need and use either. Right?
@GowthamShanmugam can you confirm the same?

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -930,3 +924,91 @@ def get_volume_alert_counts():
...
+        return devicetree
+    except Exception as ex:
+        logger.log(
+            "error",
+            NS.publisher_id,
+            {"message": "Unable to generate devicetree.err: %s" % ex}
+        )
+
+
+def find_ancestors(block_devices, device_name):
```
shtripat: Is this logic lifted from collected plugins as it was already used there?

GowthamShanmugam: no, this logic is new

GowthamShanmugam:
This PR needs: Tendrl/api/pull/451

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -930,3 +924,91 @@ def get_volume_alert_counts():
...
+        if not err:
+            out = out.encode('utf8')
+            devlist = map(
+                lambda line: dict(zip(keys, line.split(' '))),
+                out.splitlines()
+            )
+            block_devices = {}
+            for dev_info in devlist:
+                if dev_info["TYPE"] == "disk":
+                    # create new block device list
```
shtripat: Why create a new object here? Cant we add to the same object created at L#944 ?

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -930,3 +924,91 @@ def get_volume_alert_counts():
...
+                        devicetree[dev_info["MOUNTPOINT"]]["device_info"][
+                            "size"] = dev_info["SIZE"]
+                        devicetree[dev_info["MOUNTPOINT"]]["device_info"][
+                            "mountpoint"] = dev_info["MOUNTPOINT"]
+                        devicetree[dev_info["MOUNTPOINT"]]["device_info"][
+                            "type"] = dev_info["TYPE"]
+                        devicetree[dev_info["MOUNTPOINT"]]["device_info"][
+                            "size"] = dev_info["SIZE"]
+        return devicetree
+    except Exception as ex:
```
shtripat: What kind of exception we expect here? Why not catch the specific ones and take action?

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -930,3 +924,91 @@ def get_volume_alert_counts():
...
+    ancestor_details = {}
+    ancestor_details["disks"] = []
+    ancestor_details["partitions"] = []
+    while True:
+        if device_name in block_devices.keys():
+            if block_devices[device_name]["TYPE"] == "disk":
+                ancestor_details["disks"].append(
+                    block_devices[device_name]["NAME"]
+                )
+                break
```
shtripat: First device found, we break. Right?

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -930,3 +924,91 @@ def get_volume_alert_counts():
...
+        )
+
+
+def find_ancestors(block_devices, device_name):
+    # find ancestors including current dev from bottom to top
+    ancestor_details = {}
+    ancestor_details["disks"] = []
+    ancestor_details["partitions"] = []
+    while True:
+        if device_name in block_devices.keys():
```
shtripat: So if device not found in block_devices, we eventually return a blank ancestors details. Right?

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -930,3 +924,91 @@ def get_volume_alert_counts():
...
+                    block_devices[device_name]["NAME"]
+                )
+                break
+            elif block_devices[device_name]["TYPE"] == "part":
+                ancestor_details["partitions"].append(
+                    block_devices[device_name]["NAME"]
+                )
+                device_name = block_devices[device_name]["PKNAME"]
+            else:
+                # find next ancestors from next parent_name
```
shtripat: If its not partition or disk, then what value for `block_devices[device_name]["TYPE"]`?

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -930,3 +924,91 @@ def get_volume_alert_counts():
...
+            {"message": "Unable to generate devicetree.err: %s" % ex}
+        )
+
+
+def find_ancestors(block_devices, device_name):
+    # find ancestors including current dev from bottom to top
+    ancestor_details = {}
+    ancestor_details["disks"] = []
+    ancestor_details["partitions"] = []
+    while True:
```
shtripat: I feel you need a add comments to explain the logic, how you are finding the ancestors list for a device.

tendrl/gluster_integration/sds_sync/brick_device_details.py
GowthamShanmugam: yes i handled that in api PR

tendrl/gluster_integration/sds_sync/__init__.py
GowthamShanmugam: I have tested the scenarios like creating brick from single disks and creating single brick from multiple disks. Also creating brick from single disk and mount that inside already mounted brick's mount point, and creating brick from multiple disks and mount that inside already mounted brick's mount point.

tendrl/gluster_integration/sds_sync/__init__.py
GowthamShanmugam: I am not creating new object here, I am iterating devlist and populating everything inside block_devices only.

tendrl/gluster_integration/sds_sync/__init__.py
GowthamShanmugam: i will replace this with specific exception

tendrl/gluster_integration/sds_sync/__init__.py
GowthamShanmugam: yes,
let's take example lsblk output like : 

vdb                                 252:16   0  256M  0 disk 
├─GLUSTER_vg1-GLUSTER_pool1_tmeta   253:2    0    4M  0 lvm  
│ └─GLUSTER_vg1-GLUSTER_pool1-tpool 253:7    0  248M  0 lvm  
│   ├─GLUSTER_vg1-GLUSTER_pool1     253:9    0  248M  0 lvm  
│   └─GLUSTER_vg1-GLUSTER_lv1       253:11   0  248M  0 lvm  /gluster/brick1
└─GLUSTER_vg1-GLUSTER_pool1_tdata   253:3    0  248M  0 lvm  
  └─GLUSTER_vg1-GLUSTER_pool1-tpool 253:7    0  248M  0 lvm  
    ├─GLUSTER_vg1-GLUSTER_pool1     253:9    0  248M  0 lvm  
    └─GLUSTER_vg1-GLUSTER_lv1       253:11   0  248M  0 lvm  /gluster/brick1

this function is called when the mount point is found so here mount point is  /gluster/brick1, then it will loop using its parent name "GLUSTER_vg1-GLUSTER_pool1-tpool" (it will loop using kernel name of this), this is also not "disk" so again loop using its parent name "GLUSTER_vg1-GLUSTER_pool1_tmeta"  this is also not disk again loop using its parent name "vdb" now we found disk then it will break;


if you see here there are two "/gluster/brick1", but both will give the same disk information. second will override information which we collected earlier because both are the same mount point. overriding previously collected info not a problem I feel.

tendrl/gluster_integration/sds_sync/__init__.py
GowthamShanmugam: if any mount point present then only we are calling a function to find its an ancestor, so definitely ancestor will present.

if any disk mounted without partition  like:
vda                                 252:0    0   41G  0 disk  /boot

also, consider itself ancestor. 

onemore example: 
vda                                 252:0    0   41G  0 disk  
├─vda1                              252:2    0    1G  0 part /boot

this will give a result like {"/boot": "disks": [vda], "partitions": [vda1]}
for vda1 it will cosider itself also ancestor and added into partitions.

tendrl/gluster_integration/sds_sync/__init__.py
GowthamShanmugam: it might be lvm, mpath or something else but we only need the find disks and partitions of block device which have a mount point. it can be any type, we need to find an ancestor of that block device. if that block device also a disk or partitions then consider that block device also its ancestor.

tendrl/gluster_integration/sds_sync/__init__.py
GowthamShanmugam: i will add

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: ok

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: Ack. Can you explain the same using comments in code?

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: ok

tendrl/gluster_integration/sds_sync/__init__.py

tendrl/gluster_integration/sds_sync/__init__.py
shtripat: ok

shtripat: I need all this information in the code as comments

tendrl/gluster_integration/sds_sync/__init__.py
```diff
@@ -930,3 +925,128 @@ def get_volume_alert_counts():
                                          'alert_count': 0
                                          }
     return alert_counts
+
+
+def get_device_tree():
+    """{
```
shtripat: Is this sample output from the function?

shtripat: I feel you didnt get the point I meant. I wanted to explain the logic how the ancestors list is figured out.

shtripat:
Check and fix build issues.

shtripat:
@GowthamShanmugam please rebase and fix the coverage issue.

shtripat:
LGTM. Please rebase

Issue #705, opened 2018-10-22
Volume Utilization calculation not happening for all volumes when all bricks of any one volume are down


umangachapagain:
Bugzilla : 1641413

Created a gluster cluster with 3 volumes and manually brought down all bricks in one of the volumes. After import, volume utilization calculation does not happen for all volumes.

umangachapagain:
Fixed with Tendrl/gluster-integration/pull/706

PR #706, opened 2018-10-22
Fixed Volume Utilization calculation for other volumes

umangachapagain:
Fixed Volume Utilization calculation for other volumes when all the bricks in one of the volumes are down.

Bugzilla-id : 1641413
tendrl-bug-id: Tendrl/gluster-integration#705

Signed-off-by: Umanga Chapagain <chapagainumanga@gmail.com>

### DIFF
--- a/tendrl/gluster_integration/sds_sync/utilization.py
+++ b/tendrl/gluster_integration/sds_sync/utilization.py
@@ -16,8 +16,8 @@ def sync_utilization_details(volumes):
                 "error",
                 NS.publisher_id,
                 {
-                    "message": "Volume: %s is stopped."
-                               "Cannot get utilization data" % volume.name
+                    "message": "Volume: %s is stopped. "
+                               "Cannot get utilization data." % volume.name
                 }
             )
             continue
@@ -39,28 +39,72 @@ def sync_utilization_details(volumes):
     )
     out = ''
     try:
+        # 'out' is a dictonary with keys 'error' and 'success', each of
+        # which holds 'volume' and it's 'utilization' or 'error' respectively
+        # as key-value pairs.
+        # {
+        # "success": {
+        #     "volume1": {"total": 3243243, "used": 45435435, ...},
+        #     "volume2": {"total": 3243243, "used": 45435435, ...},
+        #     ....
+        # },
+        # "error": {
+        #     "volume3": "error message",
+        #     "volume4": "error message",
+        #     ....
+        # }
+        # }
+        # 'err' stores generic stderr messages for logging purpose
         out, err = cmd.communicate()
+
         util_det = json.loads(out)
-        for k, v in util_det.iteritems():
-            volume = voldict[k]
-            volume.usable_capacity = int(v['total'])
-            volume.used_capacity = int(v['used'])
-            volume.pcnt_used = str(v['pcnt_used'])
-            volume.total_inode_capacity = int(v['total_inode'])
-            volume.used_inode_capacity = int(v['used_inode'])
-            volume.pcnt_inode_used = str(v['pcnt_inode_used'])
-            volume.save()
-            cluster_used_capacity += volume.used_capacity
-            cluster_usable_capacity += volume.usable_capacity
-    except(KeyError, TypeError, ValueError, AttributeError):
+        if err != '':
+            logger.log(
+                "error",
+                NS.publisher_id,
+                {'message': 'Error : %s\n' % err}
+            )
+        if util_det:
+            if 'error' in util_det.keys():
+                logger.log(
+                    "error",
+                    NS.publisher_id,
+                    {'message': 'Error getting utilization of some '
+                     'volumes\nError: %s\n' % util_det['error']}
+                )
+
+            if 'success' in util_det.keys():
+                for k, v in util_det['success'].iteritems():
+                    try:
+                        volume = voldict[k]
+                        volume.usable_capacity = int(v['total'])
+                        volume.used_capacity = int(v['used'])
+                        volume.pcnt_used = str(v['pcnt_used'])
+                        volume.total_inode_capacity = int(v['total_inode'])
+                        volume.used_inode_capacity = int(v['used_inode'])
+                        volume.pcnt_inode_used = str(v['pcnt_inode_used'])
+                        volume.save()
+                        cluster_used_capacity += volume.used_capacity
+                        cluster_usable_capacity += volume.usable_capacity
+                    except(KeyError, TypeError, ValueError,
+                           AttributeError) as e:
+                        logger.log(
+                            "error",
+                            NS.publisher_id,
+                            {'message': "Error parsing utilization data "
+                             "of volume '%s'\nResult: %s\nError: %s\n"
+                             % (k, v, e)}
+                        )
+    except Exception as e:
         logger.log(
             "error",
             NS.publisher_id,
             {
-                "message": "Error getting utilization of "
-                "volumes %s . result: %s . err: %s" % (volnames, out, err)
+                "message": "Error getting volume utilization.\n"
+                "Error: %s\n" % e
             }
         )
+
     cluster_pcnt_used = 0
     if cluster_usable_capacity > 0:
         cluster_pcnt_used = (
--- a/tendrl/gluster_integration/sds_sync/vol_utilization.py
+++ b/tendrl/gluster_integration/sds_sync/vol_utilization.py
@@ -26,10 +26,14 @@ def showVolumeUtilization(vnames):
         if volume != "":
             try:
                 data = gfapi.getVolumeStatvfs(volume)
-            except gfapi.GlusterLibgfapiException:
-                sys.stderr.write("CRITICAL: Failed to get the "
-                                 "Volume Utilization Data\n")
-                sys.exit(-1)
+            except gfapi.GlusterLibgfapiException as e:
+                if 'error' not in ret_dict.keys():
+                    ret_dict['error'] = {}
+                ret_dict['error'][volume] = \
+                    {'message': "CRITICAL: Failed to "
+                                "get the Volume Utilization Data."
+                                "Error: %s" % e}
+                continue
             volumeCapacity = computeVolumeStats(data)
             # total size in KB
             total_size = volumeCapacity['sizeTotal'] / BYTES_IN_KB
@@ -41,7 +45,9 @@ def showVolumeUtilization(vnames):
             used_inode = data.f_files - data.f_ffree
             total_inode = data.f_files
             pcnt_inode_used = (float(used_inode) / total_inode) * 100
-            ret_dict[volume] = {
+            if 'success' not in ret_dict.keys():
+                    ret_dict['success'] = {}
+            ret_dict['success'][volume] = {
                 'total': total_size,
                 'free': free_size,
                 'used': used_size,
@@ -50,7 +56,8 @@ def showVolumeUtilization(vnames):
                 'used_inode': used_inode,
                 'pcnt_inode_used': pcnt_inode_used
             }
-    print (json.dumps(ret_dict))
+
+    print(json.dumps(ret_dict))
 
 
 def parse_input():
--- a/tendrl/gluster_integration/objects/global_details/__init__.py
+++ b/tendrl/gluster_integration/objects/global_details/__init__.py
@@ -20,8 +20,8 @@ class GlobalDetails(objects.BaseObject):
         self.value = 'clusters/{0}/GlobalDetails'
 
     def render(self):
-        self.value = self.value.format(self.integration_id or
-                                       NS.tendrl_context.integration_id)
+        self.value = self.value.format(
+            self.integration_id or NS.tendrl_context.integration_id)
         return super(GlobalDetails, self).render()
 
     def save(self, update=True, ttl=None):
--- a/tendrl/gluster_integration/sds_sync/__init__.py
+++ b/tendrl/gluster_integration/sds_sync/__init__.py
@@ -83,8 +83,8 @@ class GlusterIntegrationSdsSyncStateThread(sds_sync.SdsSyncThread):
                 _cluster = NS.tendrl.objects.Cluster(
                     integration_id=NS.tendrl_context.integration_id
                 ).load()
-                if (_cluster.status == "importing" and
-                    _cluster.current_job['status'] == 'failed') or \
+                if (_cluster.status == "importing" and (
+                    _cluster.current_job['status'] == 'failed')) or \
                     _cluster.status == "unmanaging" or \
                     _cluster.status == "set_volume_profiling":
                     continue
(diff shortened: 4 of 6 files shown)


umangachapagain:
@shtripat @GowthamShanmugam  Please review.

shtripat:
Fix the issue at https://travis-ci.org/Tendrl/gluster-integration/jobs/444762365. Otherwise looks fine.

GowthamShanmugam:
give tendrl-bug-id like Tendrl/gluster-integration#705

umangachapagain:
@GowthamShanmugam  @shtripat  Please review the new changes related to Error Parsing.

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -40,27 +40,53 @@ def sync_utilization_details(volumes):
...
-            volume.total_inode_capacity = int(v['total_inode'])
-            volume.used_inode_capacity = int(v['used_inode'])
-            volume.pcnt_inode_used = str(v['pcnt_inode_used'])
-            volume.save()
-            cluster_used_capacity += volume.used_capacity
-            cluster_usable_capacity += volume.usable_capacity
-    except(KeyError, TypeError, ValueError, AttributeError):
+        err_dict = json.loads(err)
+
+        if err_dict != '':
```
shtripat: I feel err_dict is a dictionary and if so why we check `if err_dict != ''` ?
Better we should have check like `if bool(err_dict):`

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -40,27 +40,53 @@ def sync_utilization_details(volumes):
...
+            for k, v in err_dict.iteritems():
+                if k in voldict.keys():                    
+                    logger.log(
```
shtripat: This will log error for all the failed volumes one by one. Shouldn't we merge all the failed volumes in one error and log?
Also this happens every sync time and few volumes keep failing always you would hog the log files this way.

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -40,27 +40,53 @@ def sync_utilization_details(volumes):
...
+                        "error",
+                        NS.publisher_id,
+                        {
+                            "message": "Error getting utilization of '%s'. Error: %s" %(k, v)
+                        }
+                    )
+        
+        if util_det != '':
```
shtripat: Same as above. Use `if bool(util_det):`

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -40,27 +40,53 @@ def sync_utilization_details(volumes):
...
+                    volume.save()
+                    cluster_used_capacity += volume.used_capacity
+                    cluster_usable_capacity += volume.usable_capacity
+        
+                except(KeyError, TypeError, ValueError, AttributeError) as e:
+                    logger.log(
+                        "error",
+                        NS.publisher_id,
+                        {
+                            "message": "Error getting utilization of "
```
shtripat: Its actually invalid volume utilization data because its mostly talking about parsing issues from dictionary, so we should frame the message accordingly.

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -40,27 +40,53 @@ def sync_utilization_details(volumes):
...
+                        }
+                    )
+    except:
         logger.log(
             "error",
             NS.publisher_id,
             {
-                "message": "Error getting utilization of "
-                "volumes %s . result: %s . err: %s" % (volnames, out, err)
+                "message": "Error getting utilization. Result: %s" % out
```
shtripat: `Error getting volume utilization`

umangachapagain:
@shtripat @GowthamShanmugam @anmolsachan

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -40,27 +40,52 @@ def sync_utilization_details(volumes):
     out = ''
     try:
         out, err = cmd.communicate()
+
```
shtripat: Add a comment before `cmd.communicate()` call and explain the structure of `out` expected there.

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -40,27 +40,52 @@ def sync_utilization_details(volumes):
...
-            cluster_usable_capacity += volume.usable_capacity
-    except(KeyError, TypeError, ValueError, AttributeError):
+        if err != '':
+            logger.log(
+                "error",
+                NS.publisher_id,
+                {'message' : 'Error : %s\n' %err}
+            )
+        if bool(util_det):
+            if util_det['error']:
```
shtripat: If the key `error` not there it will throw exception so iys better to check if the key exists using `util_det.has_key`

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -40,27 +40,52 @@ def sync_utilization_details(volumes):
...
+                logger.log(
+                    "error",
+                    NS.publisher_id,
+                    {'message' : 'Error getting utilization of some volumes\nError: %s\n' % util_det['error']}
+                )
+
+            if util_det['success']:
```
shtripat: Same as above

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -40,27 +40,52 @@ def sync_utilization_details(volumes):
...
+                                {'message' : "Error parsing utilization data of volume '%s'\nResult: %s\nError: %s\n" % (k, v, e)}
+                            )
+    except:
         logger.log(
             "error",
             NS.publisher_id,
             {
-                "message": "Error getting utilization of "
-                "volumes %s . result: %s . err: %s" % (volnames, out, err)
+                "message": "Error getting volume utilization.\nResult: %s\n" % out
```
shtripat: `out` might not be populated in some cases of exception I feel, so is it good to add this to log here?

tendrl/gluster_integration/sds_sync/vol_utilization.py
```diff
@@ -21,15 +21,18 @@ def computeVolumeStats(data):
 
 def showVolumeUtilization(vnames):
     volumes = vnames.split(',')
-    ret_dict = {}
+    ret_dict = {'error' : [], 'success' : []}
```
shtripat: Why error and success have to be list. I see at line#30 you create a dict with `volume` as key and error message as value. So ideally with this logic `ret_dict["error"]` will be a list of dictionaries with one keyvalue pair. Thats not clean I feel.

tendrl/gluster_integration/sds_sync/vol_utilization.py
```diff
@@ -41,15 +44,17 @@ def showVolumeUtilization(vnames):
...
             pcnt_inode_used = (float(used_inode) / total_inode) * 100
-            ret_dict[volume] = {
-                'total': total_size,
-                'free': free_size,
-                'used': used_size,
-                'pcnt_used': vol_utilization,
-                'total_inode': total_inode,
-                'used_inode': used_inode,
-                'pcnt_inode_used': pcnt_inode_used
+            success = { volume : {
```
shtripat: Same as for `error` I would except `ret_det` to be a dict with two keys `error` and `success` and their values should be dicts with volume names as keys and their values would be utilization details in case of success and error message in case of error.

tendrl/gluster_integration/sds_sync/utilization.py
umangachapagain: I have defined ret_dict to always have 'error' and 'success' as a key, which gets parsed to json as util_det. So, I think it will always have that key?

tendrl/gluster_integration/sds_sync/vol_utilization.py
umangachapagain: dict created at line #30 is a single volume related error dict that has volume as key and error as message. here the need is to have multiple such entries in case of errors in multiple volume. for this list seemed favorable. in case we go with dict as value to key - 'error', we end up creating multiple levels of nested dict. With dict as list items we have flexibility to store each volume related information in a simpler and easily accessible way. 
Also, if I try to assign the dict at line #30 to 'error', new dict item will replace the old one and we will only have a single entry in the end i.e. one value for one key.
Am I thinking right or am I complicating it? 
@anmolsachan what do you say?

tendrl/gluster_integration/sds_sync/vol_utilization.py
umangachapagain: this is what is happening.
error dict created at line#30 and success dict created at line#47 are assigned to the error and success list respectively in the ret_dict.

tendrl/gluster_integration/sds_sync/utilization.py
umangachapagain: will do that.

tendrl/gluster_integration/sds_sync/utilization.py
umangachapagain: I think there are only 2 cases(non ideal) to reach this exception.
1. when vol_utilization fails to execute and 'out' is empty - in this case I feel rather than just telling 'Error getting volume utilization', it is cleaner to say that we got an empty result which suggests that vol_utilization is failing.
2. when vol_utilization dumps invalid json to stdout - here 'out' would be populated but can not be loaded as json to util_det. So, I think it is clearer to show that we are getting invalid json which is causing this exception.

overall, it is about perception I believe. Only time this might add value is during tracing the logs.

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -39,28 +39,68 @@ def sync_utilization_details(volumes):
...
-            cluster_usable_capacity += volume.usable_capacity
-    except(KeyError, TypeError, ValueError, AttributeError):
+        if err != '':
+            logger.log(
+                "error",
+                NS.publisher_id,
+                {'message' : 'Error : %s\n' %err}
+            )
+        if bool(util_det):
+            if bool(util_det['error']):
```
anmolsachan: A better way would be :
```
if util_det:
    if 'error in' util_det :
```

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -39,28 +39,68 @@ def sync_utilization_details(volumes):
...
+                logger.log(
+                    "error",
+                    NS.publisher_id,
+                    {'message' : 'Error getting utilization of some volumes\nError: %s\n' % util_det['error']}
+                )
+
+            if bool(util_det['success']):
```
anmolsachan: same here:
 if 'error' in util_det :

tendrl/gluster_integration/sds_sync/utilization.py
anmolsachan: I would agree with shubhendu here, if out can have multiple values or empty string, then checking those things with if conditions is better and then log your messages accordingly. A message with " Result : " wont look good. Being explicit with exact error and reason would be better.

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -71,4 +96,4 @@ def sync_utilization_details(volumes):
         used_capacity=int(cluster_used_capacity),
         usable_capacity=int(cluster_usable_capacity),
         pcnt_used=str(cluster_pcnt_used)
-    ).save()
+    ).save()
```
anmolsachan: Add a new line.

tendrl/gluster_integration/sds_sync/vol_utilization.py
```diff
@@ -70,4 +70,4 @@ def main():
 
 
 if __name__ == '__main__':
-    main()
+    main()
```
anmolsachan: add a line in end

anmolsachan:
Resolve Codacy, pep8 issues. Also fix related test cases. Add test cases in a separate commit. Also it would be better to squash your code changes into one single commit.

tendrl/gluster_integration/sds_sync/utilization.py
umangachapagain: `if 'error' in util_det :`

won't this always return TRUE because 'error' key is always present, it's value might be empty.

tendrl/gluster_integration/sds_sync/utilization.py
umangachapagain: won't this always return TRUE because 'success' key is always present, it's value might be empty.

tendrl/gluster_integration/sds_sync/utilization.py
umangachapagain: I will just remove 'Out' and keep the log generic?
another problem is, Codacy expects me to not have a global `except` which is not possible here as we don't know what specific exceptions would be thrown?

tendrl/gluster_integration/sds_sync/utilization.py
```diff
@@ -39,28 +39,72 @@ def sync_utilization_details(volumes):
...
+        if util_det:
+            if 'error' in util_det.keys():
+                logger.log(
+                    "error",
+                    NS.publisher_id,
+                    {'message': 'Error getting utilization of some '
+                     'volumes\nError: %s\n' % util_det['error']}
+                )
+
+            if 'success' in util_det.keys():
```
anmolsachan: will work but .keys() is not required.

GowthamShanmugam: it is better to check with .keys(), because here success is a key only, so checking with .keys() is actually safe

GowthamShanmugam:
LGTM

tendrl/gluster_integration/sds_sync/vol_utilization.py
```diff
@@ -26,10 +26,14 @@ def showVolumeUtilization(vnames):
         if volume != "":
             try:
                 data = gfapi.getVolumeStatvfs(volume)
-            except gfapi.GlusterLibgfapiException:
-                sys.stderr.write("CRITICAL: Failed to get the "
-                                 "Volume Utilization Data\n")
-                sys.exit(-1)
+            except gfapi.GlusterLibgfapiException as e:
+                if 'error' not in ret_dict.keys():
```
shtripat: Declare ret_dict above at line#24 as `ret_dict = {"error": {}, "success": {}}`.
So you wont have to check if the key exists or not and creation of it.

GowthamShanmugam: initially, he declared like this, I told him to change because in utilization.py file he does need to worry about the error.he can proceed with success. I feel it is better to create error only any problem issue raise.

shtripat: Ack. Fine.
If verified with both happy path and with problematic volumes, good to go.

anmolsachan:
LGTM

tendrl/gluster_integration/tests/test_volume_utilization.py
```diff
@@ -10,11 +10,11 @@
 
 @mock.patch(
     'subprocess.Popen.communicate',
-    mock.Mock(return_value=('{"vol1": {"total": "20000", "used": "5000",\
+    mock.Mock(return_value=('{"success":{"vol1": {"total": "20000", "used": "5000",\
```
shtripat: Remove these fixes from here as its part of other PR already

